PR #11, opened 2016-04-24
Add build scripts

ibuildthecloud:
Depends on https://github.com/rancher/sparse-tools/pull/10

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,29 +1,6 @@
-# Compiled Object files, Static and Dynamic libs (Shared Objects)
-*.o
-*.a
-*.so
-ssync
+/.dapper
+/bin
+*.swp
+/.trash-cache
 
-# test data files
-*.bar
-
-# Folders
-_obj
-_test
 .vscode
-
-# Architecture specific extensions/prefixes
-*.[568vq]
-[568vq].out
-
-*.cgo1.go
-*.cgo2.c
-_cgo_defun.c
-_cgo_gotypes.go
-_cgo_export.*
-
-_testmain.go
-
-*.exe
-*.test
-*.prof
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -1,29 +1,4 @@
-# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
kp6, line 1: any reason you had  to clobber the whole thing?

ibuildthecloud, line 1: No, I didn't know you were partial to it, looked like a default generated gitignore from github.

ibuildthecloud, line 1: Do you want me to keep the original one?

kp6, line 1: Just .vscode line

Kirill.

-----Original Message-----

ibuildthecloud:
@kp6 I update the gitignore.  I'd like to get this PR merged if we can and then start working on fixing the issues from it.  The first problem I run into is that test won't pass because of aufs.  For this to work all scratch files need to be created in a known location or temp folder.  `ioutil.TempDir` is typicallly the easiest approach.

kp6:
@ibuildthecloud 
`TempDir` certainly makes sense for most of the tests and I'l move the location for those.

But there are some stress/benchmark tests that require 10GB+ of space and real disk device to produce sensible results. 
`Current dir` worked for the latter so far. 
Any suggestions on to how to deal with those?

kp6:
... never mind the above.
@ibuildthecloud 
here: https://github.com/rancher/sparse-tools/pull/25

ibuildthecloud:
@kp6 The build is working now, the only issue is that that current code doesn't make `go fmt`.  So seems like we could merge and then you can address the formatting changes in a different PR.

kp6:
@ibuildthecloud ok

PR #28, opened 2016-05-04
build: fixed a few minor broken refs, etc.

kp6:
there's also some cleanups and more tests


kp6:
ok, the build is green now ;-)

PR #29, opened 2016-05-04
ssync: collision retry

kp6:
depends on PR #27


kp6:
I've been running 'round the clock random 10GB volume sync tests for 10 days: no issues found

PR #30, opened 2016-05-08
Move files around to make it easy to embed sfold/ssync in other binaries

ibuildthecloud:
This just moves

```
ssync/main.go => cli/ssync/main.go
sfold/main.go => cli/sfold/main.go
```

And then exposes a public `Main()` method that can be embedded.  The github diff just makes it look like a lot more is going on here.


kp6:
cool, thanks. 
A few notes: 
- A binary file got accidentally into the PR and should be removed: sfold/sfold
-  I put the binary files into `.gitignore` some time back, but I wonder if there's a more sensible way to do it instead of adding every single go binary?

ibuildthecloud:
I typically have build scripts that dump the binaries in `bin/` and then I just ignore that folder.

I'll remove the binary file.

ibuildthecloud:
I removed the binary file

PR #31, opened 2016-05-24
SSYNC: Added support for unaligned (4K) files 

kp6:
@ibuildthecloud


kp6:
@ibuildthecloud no change in the API: the buffered I/O is chosen automatically for unaligned files

ibuildthecloud:
I tested it against https://github.com/rancher/longhorn/pull/97 and it resolves the failing tests.

PR #43, opened 2016-06-28
simplify ssync client and server

jimengliu:
simplify ssync client and server.

@yasker


yasker:
It doesn't compile...

jimengliu:
sorry, I messed up this git fork with 2 users(jimeng and jimengliu) checking in from 2 different git clones(one from my mac, one from my linux). Now it should compile and the only thing I have not fixed is the golint error for something that should be suppressed.

yasker:
It's NOT OK to have separate commits to solve conflict/compile etc. They should be merged into related commits. The commits can be modified. Just use `git push -f` to update PR after you modify a commit.

Also I saw three users: jimeng, jimengliu and Ubuntu. Please unity them to one.

And what's that commit by Kirill comes from? It seems already in the master.

The message of first commit seems misleading. You may want to merge it with the other functional commit.

cli/ssync/main.go
```diff
@@ -5,13 +5,13 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/rancher/sparse-tools/log"
```
yasker, line 8: Should get rid of `sparse-tools/log` package as well.

cli/ssync/main.go
```diff
@@ -5,13 +5,13 @@ import (
...
+	log "github.com/Sirupsen/logrus"
 	"github.com/rancher/sparse-tools/sparse"
 )
 
 func Main() {
-	defaultVerboseLogLevel := log.LevelInfo    // set if -verbose
-	defaultNonVerboseLogLevel := log.LevelWarn // set if -verbose=false
+	defaultVerboseLogLevel := log.InfoLevel    // set if -verbose
```
yasker, line 13: Verbose should be at DEBUG level, non-verbose should be at INFO level.

jimengliu:
no idea where Kirill's commit came from, it is super odd for me. I need help to get these commits straighten out. Spent most of my evening to try to figure out, but it is still not in good shape. I'm still learning how to use Git, especially when I messed things up in a way unexpected, sigh...

cli/ssync/main.go
jimengliu, line 8: Should, but there are a lot of other places using it, such as sfold and all the ssync and sfold tests etc.

sparse/client.go
```diff
@@ -79,41 +72,17 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-	layoutStream := make(chan FileInterval, 128)
-	errStream := make(chan error)
-
-	// Initiate interval loading...
-	err = loadFileLayout(abortStream, file, layoutStream, errStream)
+	err = syncFileContent(encoder, decoder, file, fileSize)
 	if err != nil {
-		log.Error("Failed to retrieve local file layout:", err)
-		return nil, err
+		log.Error("syncFileContent failed: ", err)
```
yasker, line 83: In fact it's mostly caller's job to log this error, and you can `return nil, fmt.Errorf("syncFileContent failed: ", err)` if you want to add more details.

sparse/client.go
```diff
@@ -79,41 +72,17 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
 		return nil, err
 	}
 
-	abortStream := make(chan error)
-	layoutStream := make(chan FileInterval, 128)
-	errStream := make(chan error)
-
-	// Initiate interval loading...
-	err = loadFileLayout(abortStream, file, layoutStream, errStream)
+	err = syncFileContent(encoder, decoder, file, fileSize)
```
yasker, line 81: It's more common to put this one in the same line as `if`

sparse/file_test.go
```diff
@@ -1,112 +0,0 @@
-package sparse
```
yasker, line 1: Why get rid of this test?

sparse/layout_test.go
```diff
@@ -1,192 +0,0 @@
-package sparse
```
yasker, line 1: Same here. Why remove this test?

cli/ssync/main.go
yasker, line 8: Probably it's better to remove the log in an separate commit, convert everything to logrus first.

jimengliu, line 8: Good idea, but I think we need to get this Ssync in first so that Sheng Liang can get it and then do removing of logger for all sparse-tools repo.

cli/ssync/main.go
jimengliu, line 13: will fix.

cli/ssync/main.go
yasker, line 8: In fact it will only take about half an hour to do so. I definitely prefer clean code at small expense.

jimengliu, line 8: I could do that in another new branch at the same time, let me try that.

sparse/layout_test.go
yasker, line 1: I think this one is needed?

jimengliu, line 1: we are not using RetrieveLayout() function in our new client/server. But let me try to put it back in if it compiles. We can use its test cases to generate our own tests using new functions maybe later.

yasker, line 1: We will need test cases for similar functions anyway, to make sure the file layout is correct after sync.

PR #44, opened 2016-06-28
Simpler ssync with log fixes

jimengliu:
@yasker


yasker:
Need to fix golint.

At least you can ignore the error from fiemap.go, like what did here: https://github.com/rancher/sparse-tools/blob/master/scripts/validate#L14

yasker:
Also, put log improvement in a separate PR and we can merge it first.

sparse/client.go
```diff
@@ -39,34 +26,40 @@ func SyncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
 	for retries := 1; retries >= 0; retries-- {
 		hashLocal, err = syncFile(localPath, addr, remotePath, timeout, retries > 0)
```
yasker, line 27: Better not to use named return. It's generally more error prone.

sparse/file.go
```diff
@@ -33,195 +31,56 @@ func SetupFileIO(direct bool) {
 		fileOpen = fio.OpenFile
```
yasker: It's weird to have private function pointers. It should be done using struct interface. E.g. https://github.com/rancher/longhorn/blob/master/types/types.go#L34

sparse/client.go
```diff
@@ -39,34 +26,40 @@ func SyncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-	size, err := file.Seek(0, os.SEEK_END)
+	fileSize, err := file.Seek(0, os.SEEK_END)
 	if err != nil {
 		log.Error("Failed to get size of local source file:", localPath, err)
 		return nil, err
 	}
 
-	SetupFileIO(size%Blocks == 0)
+	directIO := (fileSize%Blocks == 0)
+	SetupFileIO(directIO)
```
yasker: Will this function called outside of package? If not then it should be lower case in the first character for private function.

Also, it should use interface as I commented in `SetupFileIO`

sparse/client.go
```diff
@@ -139,334 +109,243 @@ func connect(host, port string, timeout int) net.Conn {
...
-		if err1 != nil {
-			log.Fatal("Cient protocol remote retry error:", err)
-		}
-		err1 = decoder.Decode(&statusRemote)
-		if err1 != nil {
-			log.Fatal("Cient protocol remote retry status error:", err)
-		}
-	}
-	return
+	return ack, nil
```
yasker: In any case `ack == false` is also valid? Otherwise just get rid of `ack bool` and use `err` for result.

sparse/client.go
```diff
@@ -139,334 +109,243 @@ func connect(host, port string, timeout int) net.Conn {
...
-	err = decoder.Decode(&ack)
-	if err != nil {
-		log.Fatal("Client protocol decoder error:", err)
-		return false
-	}
+func syncFileContent(encoder *gob.Encoder, decoder *gob.Decoder, file *os.File, fileSize int64) error {
 
-	return ack
-}
+	fiemap := NewFiemapFile(file)
```
yasker: Fiemap related things can be singled out as a separate function.

sparse/client.go
```diff
@@ -79,41 +72,17 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
 	decoder := gob.NewDecoder(conn)
 
-	// Use unix time as hash salt
-	salt := make([]byte, binary.MaxVarintLen64)
-	binary.PutVarint(salt, time.Now().UnixNano())
-	status := sendSyncRequest(encoder, decoder, remotePath, size, salt)
-	if !status {
-		err = fmt.Errorf("Sync request to %v failed", remotePath)
-		log.Error(err)
+	ack, err := sendSyncRequest(encoder, decoder, remotePath, fileSize)
```
yasker: You can have a object contains encoder/decoder, then you don't need to pass them in for every related functions.

sparse/file.go
```diff
@@ -33,195 +31,56 @@ func SetupFileIO(direct bool) {
...
-			}
+		if err == io.EOF {
+			log.Debug("have read at the end of file, total read: ", n)
+		} else {
+			errStr := fmt.Sprintf("File read error: %s", err)
+			log.Error(errStr)
+			return nil, fmt.Errorf(errStr)
 		}
+	} else if int64(n) != dataInterval.Len() {
+		errStr := "File read underrun"
```
yasker: Should this be continue reading?

sparse/file.go
```diff
@@ -33,195 +31,56 @@ func SetupFileIO(direct bool) {
...
-	// open file
-	file, err := fileOpen(path, os.O_WRONLY, 0)
+func HashDataInterval(file *os.File, dataInterval Interval) ([]byte, error) {
+	data, err := ReadDataInterval(file, dataInterval)
 	if err != nil {
-		log.Fatal("Failed to open file for wroting:", string(path), err)
+		return nil, err
 	}
-	defer file.Close()
+	hasher := sha1.New()
```
yasker: Better use sha512/256 for future prove(and it's faster than sha256 in 64bit cpu)

sparse/file.go
```diff
@@ -33,195 +31,56 @@ func SetupFileIO(direct bool) {
...
-		for pop, existsNext := m[pos]; len(m) > 0 && existsNext; pop, existsNext = m[pos] {
-			// pop in order range
-			log.Debug(prefix, pop)
-			orderedStream <- pop
-			delete(m, pos)
-			pos = pop.End
-		}
-	}
-	close(orderedStream)
+	return hash, nil
```
yasker: You can just use `sha1.Sum(data)`(or similar)

sparse/server.go
```diff
@@ -96,326 +153,360 @@ func serveConnection(conn net.Conn) bool {
 		var path string
 		err := decoder.Decode(&path)
 		if err != nil {
-			log.Fatal("Protocol decoder error:", err)
+			log.Error("decode path error:", err)
 			return true
 		}
+		log.Debug("got the file path: ", path)
+		//path = path + "server"
```
yasker: need to be removed

sparse/server.go
```diff
@@ -96,326 +153,360 @@ func serveConnection(conn net.Conn) bool {
...
+			if err != nil {
+				log.Error("encode ack error:", err)
+				return true
+			}
 			return true
 		}
 	}
 	// Setup close sequence
-	if directFileIO {
+	if !directFileIO {
```
yasker, line 215: Good catch!

sparse/client.go
jimengliu, line 27: True

sparse/client.go
jimengliu: yes, it should be lower case since it is not used outside the package.

sparse/client.go
jimengliu: We need to check both ack and err, if err == nil, but ack is false, sync request still failed at server side.

yasker: What I meant is, check if `ack == false` here and if it is, return error.

Then the caller of function won't need to check both variables.

sparse/file.go
jimengliu: Sure, I can use that pattern.

sparse/client.go
jimengliu: But you still need to pass around that object. Or you are thinking something else ?

sparse/file.go
jimengliu: Nice catch. This actually will never happen, the os.ReadAt will either return the len of the slice passed in, or return EOF, which is handled already. So I should remove this else if section.

sparse/client.go
yasker: Make this two fields(and any other necessary fields) as a struct, and any function call related can be written as this object's method. Trying to get something more object orientated. E.g. you can create client object.

sparse/file.go
jimengliu: sha256/512 is slower than sha1 according to online data but should be more collision resistant. Chances of having collision in our usage is not big, so speed maybe more critical. Let's leave it for sha1. What do you think ? Here is the online data:

3 times cpu for each hash algorithms:

SHA1    11.985718965530396, 11.976419925689697, 11.86873197555542
SHA256  16.662450075149536, 21.551337003707886, 17.016510963439941
SHA512  18.339390993118286, 18.11187481880188,  18.085782051086426

yasker: In fact speed is not critical here(otherwise we won't need to rewrite the ssync). Stability is. I know the chance of collision is small, but I think it make sense to make this trade off.

sparse/file.go
jimengliu: have you tried that ? That is shorter. But Go By Example uses the way here. https://gobyexample.com/sha1-hashes

sparse/server.go
jimengliu: sure

sparse/file.go
jimengliu: Sure, I am OK with it.

sparse/file.go
yasker: Don't you think it's weird pattern to have `Sum(nil)`? 

Before you ask, you should either check the go doc at https://golang.org/pkg/crypto/sha1/#Sum , or the source code at https://golang.org/src/crypto/sha1/sha1.go?s=2235:2267#L115

jimengliu: hmm, that is weird. It seems like Go By Example really needs some polish.

sparse/client.go
jimengliu: Will do.

sparse/client.go
jimengliu: That sounds reasonable. Will fix.

sparse/file.go
jimengliu: Just tried different ways doing this, but they don't look good either, due to places we need both *os.FIle and those function interfaces. I will skip doing this today for now. Maybe come back to this later discussing with you tomorrow.

yasker: Something like this?

```
type FileReaderWriter interface {
    func ReadAt(data []byte, offset int64) (int, error)
    func WriteAt(data []byte, offset int64) (int, error)
}

type BufferedFileIO struct {
    file *os.File
...
}

func NewBufferedFileIO(name string, flag int, perm os.FileMode) (*BufferedFileIO, error) 

func (b *BufferedFileIO) ReadAt(data []byte, offset int64) (int, error)
func (b *BufferedFileIO) WriteAt(data []byte, offset int64) (int, error)

type DirectFileIO struct {
...
}

func NewDirectFileIO(name string, flag int, perm os.FileMode) (*DirectFileIO, error) 

func (b *DirectFileIO) ReadAt(data []byte, offset int64) (int, error)
func (b *DirectFileIO) WriteAt(data []byte, offset int64) (int, error)
```

sparse/client.go
```diff
@@ -35,38 +21,48 @@ const connectionRetries = 5
...
 		}
 		break
 	}
-	return
+	return hashLocal, err
 }
 
 func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int, retry bool) ([]byte, error) {
-	file, err := fio.OpenFile(localPath, os.O_RDONLY, 0)
+	file, err := os.OpenFile(localPath, os.O_RDONLY, 0)
```
yasker: Update to use os.Stat() to figure out the size.

sparse/client.go
```diff
@@ -139,334 +105,233 @@ func connect(host, port string, timeout int) net.Conn {
...
 	}
 
-	var statusRemote bool
-	err = decoder.Decode(&statusRemote)
+	session.encoder.Encode(requestHeader{requestMagic, syncDone})
+	return nil
+}
+
+func (session *SyncSessionClient) sendSyncRequest(path string, size int64) error {
+	err := session.encoder.Encode(requestHeader{requestMagic, syncRequestCode})
```
yasker: I'd rather create a method in `session`, for encode/decode. 

Also, it's misleading name here because what's really happening is encode and send etc.

So it's better to be called `session.Send()` and `session.Receive()`

sparse/client.go
```diff
@@ -139,334 +105,233 @@ func connect(host, port string, timeout int) net.Conn {
...
-// prints chan codes and lengths to trace
-// - sequence and interleaving of chan processing
-// - how much of the chan buffer is used
-const traceChannelLoad = false
-
-type netXferStatus struct {
-	status    bool
-	byteCount int64
+	if reply.Code != continueSync {
+		log.Debug("got unexpected reply from server: ", reply.Code)
```
yasker: What's the result of this check? Should this result in error instead?

sparse/file.go
```diff
@@ -1,227 +1,244 @@
...
+	return file.File.WriteAt(data, offset)
 }
 
-func loadFileLayout(abortStream <-chan error, file *os.File, layoutStream chan<- FileInterval, errStream chan<- error) error {
-	size, err := file.Seek(0, os.SEEK_END)
+func (file *BufferedFileIoProcessor) getFile() *os.File {
+	return file.File
+}
+
+func (file *BufferedFileIoProcessor) fileClose() {
```
yasker: In fact you don't need these boiler-plate function if you keep the interface of FileIoProcessor aligned with os.File. Because os.File already implemented those and they will be used for FileIoProcessor.

LIke this: 

```
package main

import (
    "fmt"
    "os"
)

type FileIOProcessor interface {
    // File I/O methods for direct or bufferend I/O
    fileReadAt(data []byte, offset int64) (int, error)
    fileWriteAt(data []byte, offset int64) (int, error)
    getFile() *os.File
    Close() error
    Sync() error
    Truncate(size int64) error
    Seek(offset int64, whence int) (ret int64, err error)
    Name() string
}

type FileIO struct {
    *os.File
}

func CallFileFunction(f FileIOProcessor) string {
    return f.Name()
}

func (f *FileIO) fileReadAt(data []byte, offset int64) (int, error) {
    return 0, nil
}

func (f *FileIO) fileWriteAt(data []byte, offset int64) (int, error) {
    return 0, nil
}

func (f *FileIO) getFile() *os.File {
    return f.File
}

func main() {
    f, err := os.Open("test")
    if err != nil {
        fmt.Println(err)
    }
    fio := &FileIO{f}
    fmt.Println(CallFileFunction(fio))
}
```

Remember to distinguish ReadAt/WriteAt though.

sparse/server.go
```diff
@@ -1,18 +1,83 @@
...
+	requestMagic    requestCode = 31415926
+	syncRequestCode requestCode = 1 + iota
+	syncHole
+	syncData
+	syncDone
+)
+
+func (reqCode requestCode) String() string {
+	var s string
+	if reqCode&syncRequestCode == syncRequestCode {
```
yasker: TOTALLY WRONG!

The variable you assigned is incremental, but the way you're judging is bitwise? 

syncRequestCode = 2(`10b`), syncHole =3(`11b`), syncData = 4(`100b`), syncDone = 5(`101b`)

If `reqCode = syncHole = 3`, then `reqCode & syncRequestCode = 11b & 10b = 10b = syncRequestCode`
If `reqCode = syncDone = 5`, then `reqCode & syncData = 101b & 100b = 100b = syncData`

How can this possibly work???

I have no idea why you want to use bitwise operation here, it makes no sense

jimengliu: Good catch! I was copying online code using bitwise shifting enum definition and String() function, but then I thought there is no need to do bitwise shifting, just incrementing is good enough. So I modified that definition part but forgot the String() function. I bet debugging message has been wrong afterwards that I did not pay attention when tests were passing. 
Fixing it now.

sparse/file.go
jimengliu: I don't get what you meant. Are you saying I have to align os.File methods ? I have to almost implement all methods in os.File ?

type File
    func Create(name string) (_File, error)
    func NewFile(fd uintptr, name string) *File
    func Open(name string) (_File, error)
    func OpenFile(name string, flag int, perm FileMode) (*File, error)
    func Pipe() (r *File, w *File, err error)
    func (f *File) Chdir() error
    func (f *File) Chmod(mode FileMode) error
    func (f *File) Chown(uid, gid int) error
    func (f *File) Close() error
    func (f *File) Fd() uintptr
    func (f *File) Name() string
    func (f *File) Read(b []byte) (n int, err error)
    func (f *File) ReadAt(b []byte, off int64) (n int, err error)
    func (f *File) Readdir(n int) (fi []FileInfo, err error)
    func (f *File) Readdirnames(n int) (names []string, err error)
    func (f *File) Seek(offset int64, whence int) (ret int64, err error)
    func (f *File) Stat() (FileInfo, error)
    func (f *File) Sync() error
    func (f *File) Truncate(size int64) error
    func (f *File) Write(b []byte) (n int, err error)
    func (f *File) WriteAt(b []byte, off int64) (n int, err error)
    func (f *File) WriteString(s string) (n int, err error)

sparse/client.go
jimengliu: We shall remove it. It was just for debugging during developing the code.

sparse/client.go
jimengliu: I like this idea. Fixing it.

sparse/file.go
yasker: I've written the exactly code you need for this interface. Read the code.

PR #45, opened 2016-07-07
clean up logic for easy understanding

jimengliu:
- Error out if extents are not ordered
- change the client side sync data logic to if statements, not loop
- modify a little bit both client and server binary search comments

@yasker @sheng-liang

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -76,7 +76,9 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
 	err = session.syncFileContent(fileIo, fileSize)
 	if err != nil {
 		log.Error("syncFileContent failed: ", err)
+		return nil, err
 	}
+
 	return nil, err
 }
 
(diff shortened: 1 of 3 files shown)


sparse/sfold.go
```diff
@@ -63,25 +63,33 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
...
 
-		offset := data
-		buffer := AllocateAligned(blockSize)
-		for offset != hole {
-			// read a block from child, maybe use bufio or Reader stream
-			n, err := childFileIo.fileReadAt(buffer, offset)
-			if n != len(buffer) || err != nil {
+		batch := 32 * blockSize
+		if hole-data < int64(batch) {
+			batch = int((hole - data) * 32)
```
yasker: I don't understand what does this mean?

sparse/sfold.go
```diff
@@ -63,25 +63,33 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
 		}
 
 		// now we have a data start offset and length(hole - data)
-		// let's read from child and write to parent file block by block
+		// let's read from child and write to parent file. We read/write up to
```
yasker: It's not related to ssync change right? Please put sfold change in an separate PR.

jimengliu: yeah, it is not related to ssync. Sure I will generate another PR.

sparse/sfold.go
jimengliu: Good catch. No one will understand it, including me :-( It is a typo(*32 part).  I should have double checked this before sending out, my bad.

PR #46, opened 2016-07-08
Sfold read/write 32 blocks instead of one block

jimengliu:
Sfold read/write 32 blocks instead of one block, in order to speed things up a little.

@sheng-liang @yasker

### DIFF
--- a/sparse/sfold.go
+++ b/sparse/sfold.go
@@ -7,6 +7,10 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+const (
+	batchBlockCount = 32
+)
+
 // FoldFile folds child snapshot data into its parent
 func FoldFile(childFileName, parentFileName string) error {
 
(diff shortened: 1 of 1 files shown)


sparse/sfold.go
```diff
@@ -63,25 +63,30 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
...
+		batch := 32 * blockSize
+		buffer := AllocateAligned(batch)
+		for offset := data; offset < hole; {
+			size := batch
+			if offset+int64(size) > hole {
+				size = int(hole - offset)
+			}
+			// read a batch from child, maybe use bufio or Reader stream
+			n, err := childFileIo.fileReadAt(buffer[:size], offset)
+			if n != size || err != nil {
```
yasker: In fact fileReadAt or fileWriteAt should always error out if n != size. So no need to check that

sparse/sfold.go
```diff
@@ -63,25 +63,30 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
...
 			return err
 		}
 
-		offset := data
-		buffer := AllocateAligned(blockSize)
-		for offset != hole {
-			// read a block from child, maybe use bufio or Reader stream
-			n, err := childFileIo.fileReadAt(buffer, offset)
-			if n != len(buffer) || err != nil {
+		batch := 32 * blockSize
```
yasker: Don't use magic number here. Put 32 in a constant, or parameter probably.

jimengliu:
addressed comments.

PR #48, opened 2016-07-15
change gob to rest calls

jimengliu:
- make client and server communicate using http rest calls instead of GOB
- re-organize files and folders

@yasker @ibuildthecloud


cli/ssync/main.go
```diff
@@ -33,16 +34,17 @@ Examples:
...
-			fmt.Fprintln(os.Stderr, "Listening on", endpoint, "...")
 		} else {
 			log.SetLevel(defaultNonVerboseLogLevel)
 		}
 
-		sparse.Server(endpoint, *timeout)
+		err := rest.Server(*port)
+		if err != nil {
+			log.Fatalf("Ssync server is not started up correctly, err: %s", err)
+			os.Exit(1)
```
yasker: log.Fatalf() will call os.Exit(1) and result in process exit. Don't need this line.

cli/ssync/main.go
```diff
@@ -64,9 +66,9 @@ Examples:
 			log.SetLevel(defaultNonVerboseLogLevel)
 		}
 
-		_, err := sparse.SyncFile(srcPath, endpoint, dstPath, *timeout)
+		err := sparse.SyncFile(srcPath, endpoint, dstPath, *timeout)
 		if err != nil {
-			log.Info("ssync: error: ", err, ", exit code 1")
+			log.Errorf("ssync: error: %s, exit code 1", err)
```
yasker: Call Fatalf()

cli/ssync/main.go
```diff
@@ -33,16 +34,17 @@ Examples:
 
 	args := flag.Args()
 	if *daemon {
-		// Daemon mode
-		endpoint := sparse.TCPEndPoint{Host: "" /*bind to all*/, Port: int16(*port)}
 		if *verbose {
 			log.SetLevel(defaultVerboseLogLevel)
```
yasker: Why not just use log.DebugLevel and log.InfoLevel directly here?

sparse/file.go
```diff
@@ -45,18 +45,23 @@ func NewBufferedFileIoProcessorByFP(fp *os.File) *BufferedFileIoProcessor {
 	return &BufferedFileIoProcessor{fp}
 }
 
-func (file *BufferedFileIoProcessor) fileReadAt(data []byte, offset int64) (int, error) {
+func (file *BufferedFileIoProcessor) ReadAt(data []byte, offset int64) (int, error) {
```
yasker: Not necessary. It's os.File's interface.

sparse/file.go
```diff
@@ -45,18 +45,23 @@ func NewBufferedFileIoProcessorByFP(fp *os.File) *BufferedFileIoProcessor {
...
 	return file.File.ReadAt(data, offset)
 }
 
-func (file *BufferedFileIoProcessor) fileWriteAt(data []byte, offset int64) (int, error) {
+func (file *BufferedFileIoProcessor) WriteAt(data []byte, offset int64) (int, error) {
```
yasker: Same here.

sparse/file.go
```diff
@@ -74,7 +79,7 @@ func NewDirectFileIoProcessor(name string, flag int, perm os.FileMode, isCreate
 
 	// if file does not exist, we need to create it if asked to
 	if err != nil && len(isCreate) > 0 && isCreate[0] {
-		file, err = os.OpenFile(name, os.O_CREATE|os.O_TRUNC|syscall.O_DIRECT|flag, perm)
+		file, err = os.OpenFile(name, os.O_CREATE|syscall.O_DIRECT|flag, perm)
```
yasker, line 75: Reason of O_TRUNC or not?

jimengliu, line 75: for creation, we don't need truncate, bc it is creating a fresh 0 length file

sparse/file.go
jimengliu: will fix

sparse/file.go
jimengliu: true, will fix

cli/ssync/main.go
jimengliu: will fix

cli/ssync/main.go
jimengliu: will fix

cli/ssync/main.go
jimengliu: will fix

sparse/client.go
```diff
@@ -178,79 +131,117 @@ func (session *SyncSessionClient) syncFileContent(file FileIoProcessor, fileSize
...
+	if resp != nil && resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
+
 	return nil
 }
 
-func (session *SyncSessionClient) SyncDataInterval(file FileIoProcessor, dataInterval Interval) error {
+func (client *syncClient) syncDataInterval(file FileIoProcessor, dataInterval Interval) error {
 	const batch = 32 * Blocks
```
yasker: Magic number

sparse/client.go
```diff
@@ -264,61 +255,34 @@ func (session *SyncSessionClient) SyncDataInterval(file FileIoProcessor, dataInt
...
 			return err
 		}
+		log.Debugf("server checksum length: %d", len(serverCheckSum))
 
-		if reply.Code == sendChecksum {
-			log.Debug("server reply asking for checksum")
-
-			// calculate checksum for the data batch interval
+		serverNeedData := true
+		if len(serverCheckSum) != 0 {
```
yasker, line 260: Do you have test case for this `if` branch? Say, got checksum, then found it's different vs found it's the same.

sparse/client.go
jimengliu: will fix

sparse/file.go
```diff
@@ -219,3 +224,50 @@ func GetFiemapExtents(file FileIoProcessor) ([]Extent, error) {
 
 	return exts, nil
 }
+
+func GetFileLayout(file FileIoProcessor) ([]Interval, []Interval, error) {
+	fileSize, err := file.Seek(0, os.SEEK_END)
```
yasker: Why not just use file.Stat()?

sparse/client.go
jimengliu, line 260: yes

sparse/file.go
jimengliu: agree, stat is more straight forward, will change.

sparse/rest/handlers.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	log.Debug("receiving remote interval: ", remoteDataInterval)
+
+	data, err := ioutil.ReadAll(io.LimitReader(request.Body, remoteDataInterval.End-remoteDataInterval.Begin))
+	if err != nil {
+		errorMessage := fmt.Sprintf("ioutil.ReadAll(io.LimitReader(r.Body, end-begin)) error: %s", err)
+		server.sendErrorResponse(errorMessage, writer)
+		return
+	}
+	err = request.Body.Close()
```
yasker: I don't think you need to call request.Body.Close(), which should be already taken care of later. See: https://golang.org/pkg/net/http/#Request

sparse/layout.go
```diff
@@ -122,7 +121,8 @@ func RetrieveLayoutStream(abortStream <-chan error, file *os.File, r Interval, l
 		}
```
yasker, line 121: Not sure if it's necessary to make this function in async way.

sparse/rest/handlers.go
jimengliu: yes, handlers don't need to but http server will.

sparse/layout.go
jimengliu, line 121: maybe not, this is only used by tests created not by me, I will just leave it this way for now as long as tests are working for now.

cli/ssync/main.go
```diff
@@ -58,16 +57,15 @@ Examples:
 
 		endpoint := sparse.TCPEndPoint{Host: *host, Port: int16(*port)}
 		if *verbose {
-			log.SetLevel(defaultVerboseLogLevel)
+			log.SetLevel(log.DebugLevel)
```
yasker: This verbose check should be outside, also, don't use fmt.Fprintf(), use log instead.

sparse/file.go
ibuildthecloud, line 75: What if the file exists?

sparse/client.go
```diff
@@ -178,80 +132,121 @@ func (session *SyncSessionClient) syncFileContent(file FileIoProcessor, fileSize
...
+	return httpClient.Do(req)
+}
+
+func (client *syncClient) openServer(remotePath string) error {
+	var err error
+	var resp *http.Response
+
+	timeStart := time.Now()
+	timeStop := timeStart.Add(time.Duration(client.timeout) * time.Second)
+	for timeNow := timeStart; timeNow.Before(timeStop); timeNow = time.Now() {
```
ibuildthecloud, line 164: You can just use a ticker here and could the max number of ticks before you error out.

sparse/rest/router.go
```diff
@@ -0,0 +1,19 @@
...
+//NewRouter creates and configures a mux router
+func NewRouter(server *SyncServer) *mux.Router {
+
+	// API framework routes
+	router := mux.NewRouter().StrictSlash(true)
+
+	// Application
+	router.HandleFunc("/v1-ssync/getChecksum", server.getChecksum).Methods("GET")
+	router.HandleFunc("/v1-ssync/open", server.open).Methods("GET")
+	router.HandleFunc("/v1-ssync/close", server.close).Methods("GET")
```
ibuildthecloud: close, sendHole should be POST.  GET is idempotent, POST is not.

sparse/rest/server.go
```diff
@@ -0,0 +1,27 @@
...
+
+type SyncServer struct {
+	filePath      string
+	holeIntervals []sparse.Interval
+	dataIntervals []sparse.Interval
+	fileIo        sparse.FileIoProcessor
+}
+
+// TestServer daemon serves only one connection for each test then exits
+func TestServer(addr sparse.TCPEndPoint, timeout int) {
```
ibuildthecloud: `sparse.TCPEndpoint` is a pretty useless data structure.  Just pass around an `address` string.

sparse/rest/handlers.go
```diff
@@ -0,0 +1,226 @@
...
+	} else {
+		log.Debugf("remoteDataInterval %s is not within server.dataIntervals, send zero length checksum", remoteDataInterval)
+	}
+	log.Debugf("local checksum length: %d", len(checksum))
+
+	fmt.Fprint(writer, checksum)
+}
+
+func (server *SyncServer) writeData(writer http.ResponseWriter, request *http.Request) {
+	log.Debug("writeData called")
```
ibuildthecloud: Please don't log "method called" log statements.  There has got to be something more useful you could log, like arguments.  Or don't log at all.

sparse/rest/server.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/rancher/sparse-tools/sparse"
+)
+
+type SyncServer struct {
+	filePath      string
+	holeIntervals []sparse.Interval
```
ibuildthecloud: Can we remove all state?  I don't think this data really needs to be kept.

ibuildthecloud: Just to clarify you can keep the fileIo reference.  Just the layout info doesn't seem needed.

sparse/rest/handlers.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	// get filePath
+	queryParams := request.URL.Query()
+	if len(queryParams["filePath"]) == 0 {
+		errorMessage := fmt.Sprintf("server.open failed, queryParams filePath not exist")
+		server.sendErrorResponse(errorMessage, writer)
+		return
+	}
+	filePath := queryParams["filePath"][0]
```
yasker: Get rid of filepath here, should be passed in through CLI.

As @ibuildthecloud pointed out, there is also security risk for server to write into any file.

cli/ssync/main.go
jimengliu: fix it up the Main(), thanks for pointing it out

sparse/file.go
jimengliu, line 75: does open existing first, if it fails, do a create

sparse/client.go
jimengliu, line 164: ticker actually does not work well here, since I need to try 2 times back to back first without delay and then periodically(1 second).

sparse/rest/router.go
jimengliu: right, fixed

sparse/rest/server.go
jimengliu: true, fixed

sparse/rest/handlers.go
jimengliu: true, removed a lot of logs now

sparse/rest/server.go
jimengliu: yes, fixed

sparse/rest/handlers.go
jimengliu: fixed it.

sparse/client.go
yasker, line 164: This statement is weird. How it that different with retry with 1 second in between? I also see the comment below. Do you know why the first 1 second need to be removed to speed up test? It does sound like a trick.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-		log.Debug("send path failed: ", err)
-		return err
+func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, data []byte) (*http.Response, error) {
+	httpClient := &http.Client{Timeout: time.Duration(httpClientTimeout * time.Second)}
+
+	url := fmt.Sprintf("http://%s/v1-ssync/%s", client.remote, action)
+
+	var req *http.Request
+	if data != nil {
+		req, _ = http.NewRequest(method, url, bytes.NewBuffer(data))
```
yasker: Don't ignore error.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
+	} else {
+		req, _ = http.NewRequest(method, url, nil)
```
yasker: Same here.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-		return err
+		return fmt.Errorf("open failed, err: %s", err)
 	}
-	log.Debugf("got SyncRequest ack back: %v", ack)
-	if ack == false {
-		return fmt.Errorf("Got negative ack from server")
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
+	defer resp.Body.Close()
```
yasker: Wrong place to defer. You don't put `defer` in the last line of function, but in when you got the resource.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
 	if err != nil {
-		log.Debug("encode syncHole cmd failed: ", err)
-		return err
+		return fmt.Errorf("sendHole failed, err: %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	err = session.send(holeInterval)
+	defer resp.Body.Close()
```
yasker: Wrong place to defer.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-		log.Debugf("encode holeInterval: %s failed: %s", holeInterval, err)
-		return err
+		return nil, fmt.Errorf("getChecksum failed, err: %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	var reply replyHeader
-	err = session.receive(&reply)
+	defer resp.Body.Close()
```
yasker, line 212: Wrong place to defer.

sparse/client.go
```diff
@@ -51,207 +47,184 @@ func syncFile(localPath string, addr TCPEndPoint, remotePath string, timeout int
...
-		log.Debug("decode syncHole cmd reply failed: ", err)
-		return err
+		return fmt.Errorf("writeData failed, err: %s", err)
 	}
-	if reply.Code == continueSync {
-		log.Debug("got continueSync reply from server")
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
+	defer resp.Body.Close()
```
yasker: Wrong place.

sparse/rest/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return sparse.Interval{begin, end}, err
+}
+
+func (server *SyncServer) open(writer http.ResponseWriter, request *http.Request) {
+	// get file size
+	interval, err := server.getQueryInterval(request)
+	if err != nil {
+		errorMessage := fmt.Sprintf("server.getQueryInterval failed, err: %s", err)
```
yasker: All these `fmt.Sprintf` pattern here (and all following functions) are cumbersome. You can just let `open` call e.g. `doOpen`, which returns a `error`, then if it's not nil, then call `server.sendErrorResponse` in one call.

sparse/client.go
jimengliu: sure, need error checking

sparse/client.go
jimengliu: this is the same

sparse/client.go
jimengliu: that is right.

sparse/rest/handlers.go
jimengliu: Looks a little cumbersome, but it gives me info about at which call it fails. Here I know it fails at getQueryInterval. If using a wrapper function, then you get error info at top level function open(), ie. if you get a "Invalid Parameter" error, sometimes it is not obvious to know which sub-function call fails. In our case, you don't know if it is getQueryInterval() or NewDirectFileIoProcessor() or Truncate() fails.

yasker: I assumed you will use fmt.Errorf() in the inner function.

jimengliu: I don't like another layer of wrapping, it is not necessary in our case. How about less lines like this?

```
if err != nil {
    server.sendErrorResponse(fmt.Sprintf("server.getQueryInterval failed, err: %s", err), writer)
    return
}
```

Here it is one line shorter than original code.

yasker: It's not about one line shorter or not.

It's normal for function to return err rather than calling `server.sendErrorResponse` every time when you caught a error.

But since the http handler has no error return, so it make sense for it to call a function with error return and deal with it in one place, rather than calling `server.sendErrorResponse` every time it got a error.

The callings to `server.sendErrorResponse` should be in one place, rather than scattering in the every step of handling.

jimengliu: So it is about better coding style of error handling. I see your point now and fixed

PR #49, opened 2016-08-03
optimize ssync and fix sfold infinite loop folding issue

jimengliu:
- optimize ssync client reading file just once either for checksum, send data or both
- fix sfold infinite loop folding from pure hole child
- add a test case for sfold_test when folding pure hole child into parent

@yasker @sheng-liang

### DIFF
--- a/sparse/file.go
+++ b/sparse/file.go
@@ -163,11 +163,15 @@ func WriteDataInterval(file FileIoProcessor, dataInterval Interval, data []byte)
 	return nil
 }
 
-func HashDataInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
+func HashFileInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
 	data, err := ReadDataInterval(file, dataInterval)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


yasker:
Can you fill a bug in Longhorn for this? We also need to add a test case for longhorn part.

yasker:
Also it's better to separate the optimization part from the bug fix, as two commits. It's easier for review and backtrace.

sparse/file.go
```diff
@@ -163,11 +163,15 @@ func WriteDataInterval(file FileIoProcessor, dataInterval Interval, data []byte)
...
-func HashDataInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
+func HashFileInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
 	data, err := ReadDataInterval(file, dataInterval)
 	if err != nil {
 		return nil, err
 	}
+	return HashDataInterval(data)
+}
+
+func HashDataInterval(data []byte) ([]byte, error) {
```
yasker, line 174: Let's just call it HashData(). There is no Interval part here.

jimengliu:
sure, will split and rename.

PR #51, opened 2016-08-03
fix pure hole child folding into a parent

jimengliu:
- replace seek_data/seek_hole using fiemap to loop through data extents
- add a test case that fold pure hole into parent

@yasker @sheng-liang

### DIFF
--- a/sparse/test/testutil_test.go
+++ b/sparse/test/testutil_test.go
@@ -125,7 +125,10 @@ func checkTestSparseFile(name string, layout []FileInterval) error {
 			if len(layoutActual) != 1 {
 				return errors.New(fmt.Sprint("hole check failure at", interval))
 			}
-			if layoutActual[0] != interval {
+			// actual hole must cover interval span
+			if layoutActual[0].Kind != interval.Kind ||
+				layoutActual[0].Begin > interval.Begin ||
+				layoutActual[0].End < interval.End {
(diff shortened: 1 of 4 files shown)


sparse/test/testutil_test.go
```diff
@@ -125,7 +125,10 @@ func checkTestSparseFile(name string, layout []FileInterval) error {
 			if len(layoutActual) != 1 {
 				return errors.New(fmt.Sprint("hole check failure at", interval))
 			}
-			if layoutActual[0] != interval {
+			// actual hole must cover interval span
```
yasker, line 128: A question: why it's not exactly the same?

jimengliu, line 128: the hole/data we get back from fiemap may not start at the specified starting point, BeginOffset may be smaller than specified, EndOffset maybe larger than specified. So here we have DHH(data, hole, hole) blocks as the layout, we loop 2 times, but actual layout we get back is the same in these 2 times, it is always [1,3], not [1,2] first and [2,3] second.

The fix could either be merging the hole and don't loop 2 times, or change comparison statement as I did.

yasker, line 128: Is it valid case to have hole adjacent to another?

yasker:
I will merge this one for now, just to unblock @sangeethah . But we need to do code cleanup/refactor for sparse-tools after this merged.

PR #54, opened 2016-09-02
add sfold error logging and remove checking logic in GetFiemapRegionExts

jimengliu:
add sfold error logging for future and remove checking logic for getting fiemap exts in a region for issue #218

@yasker

### DIFF
--- a/sparse/file.go
+++ b/sparse/file.go
@@ -220,10 +220,5 @@ func GetFiemapRegionExts(file FileIoProcessor, interval Interval) ([]Extent, err
 		lastExtStart = ext.Logical
 	}
 
-	// last ext should have the FIEMAP_EXTENT_LAST set, otherwise we don't get all exts
-	if exts[len(exts)-1].Flags&FIEMAP_EXTENT_LAST == 0 {
-		return exts, fmt.Errorf("The exts returned by fiemap are not complete")
-	}
-
 	return exts, nil
 }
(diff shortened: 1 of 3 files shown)


sparse/file.go
```diff
@@ -220,10 +220,5 @@ func GetFiemapRegionExts(file FileIoProcessor, interval Interval) ([]Extent, err
 		lastExtStart = ext.Logical
 	}
 
-	// last ext should have the FIEMAP_EXTENT_LAST set, otherwise we don't get all exts
-	if exts[len(exts)-1].Flags&FIEMAP_EXTENT_LAST == 0 {
```
yasker, line 224: Why FIEMAP_EXTENT_LAST can be unset? In other way, why we have to ensure it's set before?

jimengliu: The flag is for the file's last ext, not region's last ext. Code used to get entire file's exts and we had suspicious corruption issue during stress test, it was added to ensure it was not due to incomplete exts.

yasker: Then why it won't be triggered before? Or in our local build environment?

jimengliu: tried logging file ext count out when this failed on Drone. Found that file ext count and region ext count are all equal to 1, so this should not fail no matter where tests run as long as File System implementations of this flag are the same. At this point, I have no conclusion for the weirdness yet.

yasker:
I will merge this for now. But the question remain unanswered.

PR #58, opened 2019-08-08
Report progress of sfold operation

ttpcodes:
This PR modifies the logic of the `sfold` operation to allow for the `Progress` of the operation to be reported.

As part of the call to `FoldFile`, the user must provide a `struct` that implements `FoldFileOperations`, an interface declaring a `method` named `UpdateFoldFileProgress`, that will be used to report the status of the `FoldFile` operation via `callback`.

As part of the changes to allow for `sfold` to report its `Progress`, the `coalesce` function that `FoldFile` uses has been converted to a `goroutine`, meaning that `FoldFile` no longer blocks and will immediately return once the `sfold` operation has started. It is up to the user to implement `FoldFileOperations` and respond to the `Progress` changes of the operation accordingly.

The default implementations used by `sfold` in the `CLI` and in the tests have also been modified to implement `FoldFileOperations`, waiting for a `Done` status and sending any `error` that might've occurred during the operation.

### DIFF
--- a/sparse/sfold.go
+++ b/sparse/sfold.go
@@ -1,6 +1,7 @@
 package sparse
 
 import (
+	"fmt"
 	"os"
 	"syscall"
 
@@ -8,11 +9,16 @@ import (
 )
 
 const (
-	batchBlockCount = 32
+	batchBlockCount  = 32
+	progressComplete = 100
 )
 
+type FoldFileOperations interface {
+	UpdateFoldFileProgress(progress int, done bool, err error)
+}
+
 // FoldFile folds child snapshot data into its parent
-func FoldFile(childFileName, parentFileName string) error {
+func FoldFile(childFileName, parentFileName string, ops FoldFileOperations) error {
 
 	childFInfo, err := os.Stat(childFileName)
 	if err != nil {
@@ -33,6 +39,23 @@ func FoldFile(childFileName, parentFileName string) error {
 		panic("file sizes are not equal")
 	}
 
+	go coalesce(parentFileName, childFileName, childFInfo.Size(), ops)
+
+	return nil
+}
+
+func coalesce(parentFileName, childFileName string, fileSize int64, ops FoldFileOperations) (err error) {
+	var progress int
+
+	defer func() {
+		if err != nil {
+			log.Errorf(err.Error())
+			ops.UpdateFoldFileProgress(progress, true, err)
+		} else {
+			ops.UpdateFoldFileProgress(progressComplete, true, nil)
+		}
+	}()
+
 	// open child and parent files
 	childFileIo, err := NewDirectFileIoProcessor(childFileName, os.O_RDONLY, 0)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


sparse/sfold.go
```diff
@@ -59,41 +62,67 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
...
+	if err != nil {
+		log.Errorf("could not Stat childFile filename %s, err: %v", childFileIo.Name(), err)
+		return err
+	}
+	childFileSize := childFileInfo.Size()
+
+	go func() {
+		var err error
+		var progress int
+		defer func() {
```
yasker: Use 
```
go func() (err error) {
}
```
instead so you won't need to remember to set error when return

sparse/sfold.go
```diff
@@ -38,18 +43,16 @@ func FoldFile(childFileName, parentFileName string) error {
 	if err != nil {
 		panic("Failed to open childFile, error: " + err.Error())
 	}
-	defer childFileIo.Close()
```
yasker: If get blocksize or get exts errors out below, files won't be closed.

Also, need another commit to change all those `panic` into `return error`. Now sfold is in the main process, they cannot just panic.

sparse/sfold.go
```diff
@@ -59,41 +62,67 @@ func coalesce(parentFileIo FileIoProcessor, childFileIo FileIoProcessor) error {
...
-				size = int(dataEnd - offset)
-			}
-			// read a batch from child
-			n, err := childFileIo.ReadAt(buffer[:size], offset)
 			if err != nil {
-				log.Errorf("Failed to read childFile filename: %v, size: %v, at: %v", childFileIo.Name(), size, offset)
-				return err
+				ops.UpdateFoldFileProgress(progress, true, err)
+			} else {
+				ops.UpdateFoldFileProgress(100, true, nil)
```
yasker: Use constant instead of `100`

yasker (changes requested):

ttpcodes:
- Set `error` in `coalesce` as named return parameter to avoid having to manually set `err`.
- Stored `100` as a `constant` value (`progressComplete`).
- Moved `DirectFileIoProcessor` creation into `coalesce` as part of `goroutine` to clean up `defer` logic.
- Replaced calls to `panic` with returns of `errors`.

sparse/sfold.go
```diff
@@ -33,6 +37,22 @@ func FoldFile(childFileName, parentFileName string) error {
...
+
+	return nil
+}
+
+func coalesce(parentFileName, childFileName string, ops FoldFileOperations) (err error) {
+	var progress int
+
+	defer func() {
+		if err != nil {
+			ops.UpdateFoldFileProgress(progress, true, err)
```
yasker, line 52: Since previous we have logs for each statement, it's better we keep it that way to print log when an error happens as well.

sparse/sfold.go
```diff
@@ -46,19 +66,20 @@ func FoldFile(childFileName, parentFileName string) error {
...
 	if err != nil {
 		panic("can't get FS block size, error: " + err.Error())
 	}
 	exts, err := GetFiemapExtents(childFileIo)
 	if err != nil {
-		log.Errorf("Failed to GetFiemapExtents of childFile filename: %s, err: %v", childFileIo.Name(), err)
-		return err
+		return fmt.Errorf("failed to GetFiemapExtents of childFile filename: %s, err: %v", childFileIo.Name(), err)
 	}
+	childFileInfo, err := childFileIo.Stat()
```
yasker: Size can be passed in as a parameter since the caller has the information.

yasker (changes requested):

ttpcodes:
- Any `error` encountered during `coalesce` now triggers `log.Errorf` during the `defer` call.
- `fileSize` is now passed in as a parameter to avoid having to call `Stat` on the file twice.

yasker (approved):

PR #64, opened 2019-12-27
Report progress for file sync

shuo-wu:
longhorn/longhorn#46

### DIFF
--- a/sparse/rest/handlers.go
+++ b/sparse/rest/handlers.go
@@ -14,6 +14,14 @@ import (
 	"github.com/longhorn/sparse-tools/sparse"
 )
 
+type SyncFileOperations interface {
+	UpdateSyncFileProgress(size int64)
+}
+
+type SyncFileStub struct{}
+
+func (f *SyncFileStub) UpdateSyncFileProgress(size int64) {}
+
 func (server *SyncServer) getQueryInterval(request *http.Request) (sparse.Interval, error) {
 	var interval sparse.Interval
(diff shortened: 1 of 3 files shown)


sparse/client.go
```diff
@@ -285,6 +293,7 @@ func (client *syncClient) syncDataInterval(file FileIoProcessor, dataInterval In
 			}
 		}
 		offset += batchInterval.Len()
+		ops.UpdateSyncFileProgress(size)
```
yasker: Size? Are we going to use the accumulated result?

shuo-wu: Yes.

sparse/rest/handlers.go
```diff
@@ -89,7 +99,10 @@ func (server *SyncServer) close(writer http.ResponseWriter, request *http.Reques
 
 	server.fileIo.Close()
 	log.Infof("Ssync server exit(0)")
```
shuo-wu: modify this log

sparse/rest/handlers.go
```diff
@@ -89,7 +99,10 @@ func (server *SyncServer) close(writer http.ResponseWriter, request *http.Reques
...
-	os.Exit(0)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	server.srv.Shutdown(ctx)
```
shuo-wu: directly use `Close()`

PR #65, opened 2020-04-04
add endpoint for syncing metafile

utkarshmani1997:
This commit adds an endpoint `writeMetaFile` to sync the contents
of metafile by writing the data into temp file and then renaming it
to  make contents of file consistent across crashes.

Signed-off-by: Utkarsh Mani Tripathi <utkarsh.tripathi@mayadata.io>

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


utkarshmani1997:
cc: @yasker

yasker:
I am not sure why a new endpoint is needed since we already have the ability to sync metadata files.

utkarshmani1997:
Hey @yasker  i have observed that the metadata contents were getting corrupted sometimes during crash, which required manual intervention. In other words the whole process was not atomic, using this approach we can update the metafile atomically.

yasker:
if the metadata file is corrupted, then the replica will fail, then we can just resync them during the rebuild? I don't understand how it is related to the atomic or not, or how adding a call to just sync metadata file helps. You can use the existing call to sync up the metafile as well.

utkarshmani1997:
@yasker replica will fail during open calls as json is corrupted, so you need to delete the metadata file or fix the json content to start the replica. This endpoint helps in rebuilding the metadata content from other replica atomically. For example: in the current approach if some offsets were missed during resync (due to crash) the replica will fail while open call (json unmarshalling will throw error) but if we resync using this approach, data will be written to a different file and once successful it will be renamed, this way replica will never go to bad state and will be able to recover itself without having to fix the metafile.

PR #68, opened 2020-11-10
Speed up rebuilding by getting server cheksum and calculating local checksum simultaneously

shuo-wu:
longhorn/longhorn#1948

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


sparse/client.go
```diff
@@ -241,37 +242,46 @@ func (client *syncClient) syncDataInterval(file FileIoProcessor, dataInterval In
...
 
-			1. ask server for checksum
-			2. if server send back non-zero length checksum, then calculate local checksum and compare
-			3. if server checksum sent back is zero length or comparison results differences, send data
+			1. Launch 2 goroutines to ask server for checksum and calculate local checksum simultaneously.
+			2. Wait for checksum calculation complete then compare the checksums.
+			3. Send data if the checksums are different.
 		*/
-		body, err := client.getServerChecksum(batchInterval)
+		// start := time.Now()
```
yasker: ?

sparse/client.go
```diff
@@ -241,37 +242,46 @@ func (client *syncClient) syncDataInterval(file FileIoProcessor, dataInterval In
...
+		var dataBuffer, serverCheckSum, localCheckSum []byte
+		wg := sync.WaitGroup{}
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			var body []byte
+			if body, err = client.getServerChecksum(batchInterval); err != nil {
```
yasker: Define err for each go routine. I am not sure what would be the result if two go routines assign to the same variable.

yasker (changes requested):

yasker (approved):

PR #70, opened 2021-02-26
add server directIO

cclhsu:
#### Proposed Changes ####

- add server directIO.
- fix error when send/receive file size less than block size (4096) using directIO. 
- Add/Update unit test.

#### Types of Changes ####

enhancement

#### Verification ####

- Add/Update unit test.

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/1943

#### Further Comments ####

None

Signed-off-by: Clark Hsu <clark.hsu@suse.com>

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -37,7 +37,10 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 		return err
 	}
 	fileSize := fileInfo.Size()
-	log.Infof("source file size: %d, setting up directIo: %v", fileSize, directIO)
+	if directIO && fileSize%Blocks != 0 {
+		return fmt.Errorf("Invalid directIO file block size: %v", fileSize)
+	}
+	log.Infof("source file size: %d, setting up directIO: %v", fileSize, directIO)
 
 	var fileIo FileIoProcessor
 	if directIO {
@@ -55,7 +58,7 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 
 	defer client.closeServer() // kill the server no matter success or not, best effort
 
-	err = client.syncFileContent(fileIo, fileSize)
+	err = client.syncFileContent(fileIo, fileSize, directIO)
 	if err != nil {
 		log.Errorf("syncFileContent failed: %s", err)
 		return err
@@ -64,13 +67,13 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 	return err
 }
 
-func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64) error {
+func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64, directIO bool) error {
 	exts, err := GetFiemapExtents(file)
 	if err != nil {
 		return err
 	}
 
-	err = client.openServer()
+	err = client.openServer(directIO)
 	if err != nil {
 		return fmt.Errorf("openServer failed, err: %s", err)
 	}
@@ -127,7 +130,7 @@ func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64) 
 	return nil
 }
 
-func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, data []byte) (*http.Response, error) {
+func (client *syncClient) sendHTTPRequest(method string, action string, queries map[string]string, data []byte) (*http.Response, error) {
 	httpClient := &http.Client{Timeout: time.Duration(httpClientTimeout * time.Second)}
 
 	url := fmt.Sprintf("http://%s/v1-ssync/%s", client.remote, action)
@@ -146,8 +149,9 @@ func (client *syncClient) sendHTTPRequest(method string, action string, interval
 	req.Header.Add("Accept", "application/json")
 
 	q := req.URL.Query()
-	q.Add("begin", strconv.FormatInt(interval.Begin, 10))
-	q.Add("end", strconv.FormatInt(interval.End, 10))
+	for k, v := range queries {
+		q.Add(k, v)
+	}
 	req.URL.RawQuery = q.Encode()
 
 	log.Debugf("method: %s, url with query string: %s, data len: %d", method, req.URL.String(), len(data))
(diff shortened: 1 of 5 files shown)


shuo-wu (changes requested):
~~When the sender starts to send a file, I think it should tell the server if it can use direct IO:~~
~~https://github.com/longhorn/sparse-tools/blob/c1877f4a574f57a461f7a69709defd20bc52718a/sparse/client.go#L73~~

~~https://github.com/longhorn/sparse-tools/blob/c1877f4a574f57a461f7a69709defd20bc52718a/sparse/rest/handlers.go#L51-L70~~

sparse/client.go
```diff
@@ -144,6 +147,7 @@ func (client *syncClient) sendHTTPRequest(method string, action string, interval
 	}
 
 	req.Header.Add("Accept", "application/json")
+	req.Header.Add("DirectIO", strconv.FormatBool(directIO))
```
shuo-wu: Can we put this parameter in `req.URL.Query()`?

sparse/client.go
```diff
@@ -184,12 +188,12 @@ func (client *syncClient) openServer() error {
 	return nil
 }
 
-func (client *syncClient) closeServer() {
-	client.sendHTTPRequest("POST", "close", Interval{0, 0}, nil)
+func (client *syncClient) closeServer(directIO bool) {
```
shuo-wu: It seems that the server won't check this new parameter `directIO` and actually it is no need to check it. In other words, we don't need to modify the function signature.

Same for the below `syncClient.getServerChecksum` and `syncClient.writeData`.

sparse/client.go
```diff
@@ -127,7 +130,7 @@ func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64)
 	return nil
 }
 
-func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, data []byte) (*http.Response, error) {
+func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, data []byte, directIO bool) (*http.Response, error) {
```
shuo-wu: The new parameter `directIO` is required by the caller `client.openServer(directIO)` only. Asking other callers to set this option doesn't make sense. Hence, can we modify the input like the following?
```
sendHTTPRequest(method string, action string, querys map[string]string, data []byte)
```

shuo-wu (changes requested):

sparse/client.go
cclhsu: changed

sparse/client.go
cclhsu: changed

sparse/client.go
cclhsu: changed.

cclhsu:
@shuo-wu I have updated the code according the review and integation test in https://github.com/longhorn/longhorn-engine/pull/560 has passed.

sparse/client.go
```diff
@@ -127,7 +130,7 @@ func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64)
 	return nil
 }
 
-func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, data []byte) (*http.Response, error) {
+func (client *syncClient) sendHTTPRequest(method string, action string, interval Interval, queries map[string]string, data []byte) (*http.Response, error) {
```
shuo-wu: I mean `func (client *syncClient) sendHTTPRequest(method string, action string, queries map[string]string, data []byte) (*http.Response, error)`

All callers need to put `interval.Begin` and `interval.End` in `queries` before calling this function. Then This function can blindly put all entries of `queries` in `req.URL.Query()` without `directIO, err := queries["directIO"]`

sparse/rest/handlers.go
```diff
@@ -22,6 +22,26 @@ type SyncFileStub struct{}
...
+		return directIO, err
+	}
+
+	directIO, err = strconv.ParseBool(directIOStr)
+	if err != nil {
+		err = fmt.Errorf("strconv.ParseBool(directIOStr) error: %s", err)
+		return directIO, err
+	}
+
+	return directIO, err
```
shuo-wu: I would prefer this kind of style:
```suggest
	queryParams := request.URL.Query()
	directIOStr := queryParams.Get("directIO")
	if directIOStr == "" {
		return false, fmt.Errorf("directIO does not exist in the queries: %+v", queryParams)
	}

	directIO, err := strconv.ParseBool(directIOStr)
	if err != nil {
		err = fmt.Errorf("strconv.ParseBool(directIOStr) error: %s", err)
		return false, errors.Warpf(err, "failed to parse directIO %v", directIOStr)
	}
	return directIO, nil
```

Returning `return directIO, err` in the last line is a little bit confusing.

shuo-wu (changes requested):

sparse/client.go
cclhsu: updated

sparse/rest/handlers.go
cclhsu: updated

sparse/client.go
```diff
@@ -146,23 +149,28 @@ func (client *syncClient) sendHTTPRequest(method string, action string, interval
...
-func (client *syncClient) openServer() error {
+func (client *syncClient) openServer(directIO bool) error {
 	var err error
 	var resp *http.Response
 
 	timeStart := time.Now()
 	timeStop := timeStart.Add(time.Duration(client.timeout) * time.Second)
+	queries := make(map[string]string)
+	queries["begin"] = strconv.FormatInt(0, 10)
+	queries["end"] = strconv.FormatInt(client.fileSize, 10)
```
shuo-wu, line 170: Don't need to add these 2 entries. Since the server won't invoke `getQueryInterval` after receiving `open` calls.

Same for `func (client *syncClient) closeServer() {`

sparse/rest/handlers.go
```diff
@@ -22,6 +22,23 @@ type SyncFileStub struct{}
 
 func (f *SyncFileStub) UpdateSyncFileProgress(size int64) {}
 
+func (server *SyncServer) getQueryDirectIO(request *http.Request) (bool, error) {
+	var directIO bool
+	var err error
```
shuo-wu: redundant if you modify line 35 to the following: 
```
directIO, err := strconv.ParseBool(directIOStr)
```

shuo-wu (changes requested):

sparse/client.go
cclhsu: updated

sparse/rest/handlers.go
cclhsu: updated

sparse/client.go
shuo-wu: OH, sorry. This suggestion should be applied to `closeServer` only. And you already found that.

shuo-wu (approved):
LGTM

PR #71, opened 2021-03-04
add logs for debugging snapshot (#2161)

cclhsu:
#### Proposed Changes ####

add logs for debugging snapshot to resolve ticket '[FEATURE]Add more debug logs for snapshotting' (#2161)

- add sendfile time

#### Types of Changes ####

- enhancement

#### Verification ####

Additional log like following

2021-03-04T04:10:53.970318517Z time="2021-03-04T04:10:53Z" level=info msg="Finish sending file /tmp/test256KB.tmp and take 1.644776033s"

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2161

#### Further Comments ####

Signed-off-by: Clark Hsu <clark.hsu@suse.com>
```

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -38,6 +38,8 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 	}
 	fileSize := fileInfo.Size()
 	log.Infof("source file size: %d, setting up directIo: %v", fileSize, directIO)
+	startTime := time.Now()
+	log.Infof("Start sending file %s", localPath)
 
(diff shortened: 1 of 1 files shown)


shuo-wu (changes requested):
- ~~Typically there is no need to calculate the elapsed time since there will be a timestamp for each log message.~~ It's fine to calculate the elapsed time for this function since it may take a long time. 
- `SyncFile` is related to replica rebuilding rather than snapshot. We can add logs for this function but you need to rename the commit/PR title

PR #72, opened 2021-03-25
Downgrade some logs from debug level to trace level

shuo-wu:
Otherwise, these logs will flood callers,
e.g., backing image manager.

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -154,7 +154,7 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries 
 	}
 	req.URL.RawQuery = q.Encode()
 
-	log.Debugf("method: %s, url with query string: %s, data len: %d", method, req.URL.String(), len(data))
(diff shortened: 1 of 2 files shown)


sparse/client.go
```diff
@@ -154,7 +154,7 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
 	}
 	req.URL.RawQuery = q.Encode()
 
-	log.Debugf("method: %s, url with query string: %s, data len: %d", method, req.URL.String(), len(data))
+	// log.Debugf("method: %s, url with query string: %s, data len: %d", method, req.URL.String(), len(data))
```
innobead: Can we use trace instead of commenting out? if it's helpful for the development. The same suggestion for others.

innobead (changes requested):

sparse/client.go
shuo-wu: OK. Let me try it first.

shuo-wu: Updated

innobead (approved):

PR #73, opened 2021-04-07
Fix typo

jenting:
Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>


innobead (approved):
LGTM. I think we should introduce the typo correction in the PR check :)

PR #74, opened 2021-04-13
Enable cancellation during file sync

shuo-wu:
longhorn/longhorn#2416

### DIFF
--- a/cli/ssync/main.go
+++ b/cli/ssync/main.go
@@ -1,6 +1,7 @@
 package ssync
 
 import (
+	"context"
 	"flag"
 
 	log "github.com/sirupsen/logrus"
@@ -42,7 +43,8 @@ func Main() {
 		dstPath := args[0]
 
 		ops := &rest.SyncFileStub{}
-		err := rest.Server(*port, dstPath, ops)
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		err := rest.Server(ctx, cancelFunc, *port, dstPath, ops)
 		if err != nil {
 			log.Fatalf("Ssync server failed, err: %s", err)
 		}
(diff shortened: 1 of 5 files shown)


shuo-wu:
Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/155/

innobead:
> Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/155/

regression failed?

shuo-wu:
> > Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/155/
> 
> regression failed?

13 test cases are failed in the regression result. And right now there are around 12 test cases failed in night tests. I will check if some failures are caused by this PR later.

cli/ssync/main.go
```diff
@@ -42,7 +43,8 @@ func Main() {
 		dstPath := args[0]
 
 		ops := &rest.SyncFileStub{}
-		err := rest.Server(*port, dstPath, ops)
+		ctx, cancelFunc := context.WithCancel(context.Background())
```
innobead, line 46: It seems in this case, I assume there is no need to run cancelFunc to stop the server before the process finished. The change here is just to conform to the interface change.

innobead (approved):
LGTM

cclhsu (approved):

cli/ssync/main.go
shuo-wu: Yes. And this part is no longer used by any longhorn workloads. Now it's only for manually debugging the sync feature.

shuo-wu:
Directly running the failed cases that the failure reasons are different from that of the nightly test gets passed. And the previous failure reasons seem to be not related to rebuilding/BIM. I think it's good to go.

PR #76, opened 2021-05-26
1948 replica rebuild speed up

joshimoo:
iterativly retrieve file extents 1024 at a time, these get converted to their relevant fileInterval and pushed out to a buffered channel (10 * 1024) which then gets processed concurrently by 4 worker go routines. 

Below are some test benchmarks for the local unit test, 10x improvement for the worst case 4KB blocks / 4KB holes file
This is the worst case, since in the previous implementation we would have the upfront cost of retrieving & storing all extents as well as pay the http overhead/delay price per extent.

// 120s for 1GB sync on a local server is way to slow
// 12s for 1GB parallel sync segments: 262144/11.32s data: 131072/7.30s holes: 131072/3.90s

For a dense file:
// 1GB file with 1MB data blocks segments: 91857/14.80s data: 90833/14.64s holes: 1024/0.11s
// 1GB file with 4MB data blocks segments: 88827/14.82s data: 88571/14.75s holes: 256/0.01s
// 1GB file with 32MB data blocks segments: 71577/11.92s data: 71545/11.87s holes: 32/0.00s
// 10GB file with 128MB data blocks segments: 716762/132.81s data: 716682/132.40s holes: 80/0.01s
// retrieved extents for file ssync-src-fiemap-10gb-file-dense-128MB, fileSize: 10737418240 elapsed: 130.84s, extents: 716682

extent count is based on the physical structure of the file, a file with the same layout can have different amount of extents
depending on where the filesystem stores the individual blocks, on a defragmented file syncing might reduce to workerCount * io
since we currently parallelize on segments instead of the actual data interval batched (32 * 4096) if this becomes a problem
we can also process the batches for a single data interval concurrently. 

Even if a file is extremely continuous, we would still benefit from the no preprocessing pass as well as  the 4x concurrency factor together with the (32 * 4K) data sends the overhead per send should be lower in comparison to the max sparse case above, we are also able to reuse the prior http connection now.

**Will write up test/benchmark instructions for in cluster testing tomorrow**
**This PR also contains a concurrent fold implementation that is used for snapshot coalescing**

I might also add an additional optimization for checksum calculation, after testing tomorrow.

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -2,8 +2,10 @@ package sparse
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -25,6 +27,24 @@ type syncClient struct {
 	filePath string
 	fileSize int64
 	fileIo   FileIoProcessor
+
+	httpClient *http.Client
+}
+
+func newHTTPClient() *http.Client {
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.MaxIdleConns = 100
+	t.MaxConnsPerHost = 100
+	t.MaxIdleConnsPerHost = 100
+
+	return &http.Client{
+		Timeout:   httpClientTimeout * time.Second,
+		Transport: t,
+	}
+}
+
+func newSyncClient(remote string, timeout int, filePath string, fileSize int64, fileIo FileIoProcessor) *syncClient {
+	return &syncClient{remote: remote, timeout: timeout, filePath: filePath, fileSize: fileSize, fileIo: fileIo, httpClient: newHTTPClient()}
 }
 
 const connectionRetries = 5
@@ -54,7 +74,7 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 	}
 	defer fileIo.Close()
 
-	client := &syncClient{remote, timeout, localPath, fileSize, fileIo}
+	client := newSyncClient(remote, timeout, localPath, fileSize, fileIo)
 
 	defer client.closeServer() // kill the server no matter success or not, best effort
 
@@ -68,70 +88,56 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 }
 
 func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64, directIO bool) error {
-	exts, err := GetFiemapExtents(file)
-	if err != nil {
-		return err
+	if err := client.openServer(directIO); err != nil {
+		return fmt.Errorf("openServer failed, err: %v", err)
 	}
 
-	err = client.openServer(directIO)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	syncStartTime := time.Now()
+	out, errc, err := GetFileLayout(ctx, file)
 	if err != nil {
-		return fmt.Errorf("openServer failed, err: %s", err)
+		return fmt.Errorf("failed to retrieve file layout for file %v err: %v", file.Name(), err)
 	}
 
-	var lastIntervalEnd int64
-	var holeInterval Interval
-
-	for _, e := range exts {
-		interval := Interval{int64(e.Logical), int64(e.Logical + e.Length)}
-
-		// report hole
-		if lastIntervalEnd < interval.Begin {
-			holeInterval = Interval{lastIntervalEnd, interval.Begin}
-			err := client.syncHoleInterval(holeInterval)
-			if err != nil {
-				return fmt.Errorf("syncHoleInterval %s failed, err: %s", holeInterval, err)
+	processSegment := func(segment FileInterval) error {
+		if segment.Kind == SparseHole {
+			if err := client.syncHoleInterval(segment.Interval); err != nil {
+				return fmt.Errorf("syncHoleInterval %s failed, err: %s", segment.Interval, err)
+			}
+		} else if segment.Kind == SparseData {
+			if err := client.syncDataInterval(file, segment.Interval); err != nil {
+				return fmt.Errorf("syncDataInterval %s failed, err: %s", segment.Interval, err)
 			}
 		}
-		lastIntervalEnd = interval.End
-
-		// report data
-		err := client.syncDataInterval(file, interval)
-		if err != nil {
-			return fmt.Errorf("syncDataInterval %s failed, err: %s", interval, err)
-		}
-
-		if e.Flags&FIEMAP_EXTENT_LAST != 0 {
-
-			// report last hole
-			if lastIntervalEnd < fileSize {
-				holeInterval := Interval{lastIntervalEnd, fileSize}
-
-				// syncing hole interval
-				err = client.syncHoleInterval(holeInterval)
-				if err != nil {
-					return fmt.Errorf("syncHoleInterval %s failed, err: %s", holeInterval, err)
-				}
-			}
-		}
+		return nil
 	}
 
-	// special case, the whole file is a hole
-	if len(exts) == 0 && fileSize != 0 {
-		holeInterval = Interval{0, fileSize}
-		log.Infof("The file is a hole: %s", holeInterval)
-
-		// syncing hole interval
-		err := client.syncHoleInterval(holeInterval)
-		if err != nil {
-			return fmt.Errorf("syncHoleInterval %s failed, err: %s", holeInterval, err)
-		}
+	const WorkerCount = 4
+	errorChannels := []<-chan error{errc}
+	for i := 0; i < WorkerCount; i++ {
+		errorChannels = append(errorChannels, processFileIntervals(ctx, out, processSegment))
 	}
 
-	return nil
+	// the below select will exit once all error channels are closed, or a single
+	// channel has run into an error, which will lead to the ctx being cancelled
+	mergedErrc := mergeErrorChannels(ctx, errorChannels...)
+	select {
+	case err = <-mergedErrc:
+		break
+	}
+
+	log.Debugf("finished sync for file: %v size: %v elapsed: %.2fs",
+		file.Name(), fileSize, time.Now().Sub(syncStartTime).Seconds())
+	return err
 }
 
 func (client *syncClient) sendHTTPRequest(method string, action string, queries map[string]string, data []byte) (*http.Response, error) {
-	httpClient := &http.Client{Timeout: time.Duration(httpClientTimeout * time.Second)}
+	httpClient := client.httpClient
+	if httpClient == nil {
+		httpClient = newHTTPClient()
+	}
 
 	url := fmt.Sprintf("http://%s/v1-ssync/%s", client.remote, action)
 
@@ -184,17 +190,26 @@ func (client *syncClient) openServer(directIO bool) error {
 	if err != nil {
 		return fmt.Errorf("open failed, err: %s", err)
 	}
+
+	// drain the buffer and close the body
+	_, _ = io.Copy(ioutil.Discard, resp.Body)
+	_ = resp.Body.Close()
+
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	resp.Body.Close()
 
 	return nil
 }
 
 func (client *syncClient) closeServer() {
 	queries := make(map[string]string)
-	client.sendHTTPRequest("POST", "close", queries, nil)
+	resp, err := client.sendHTTPRequest("POST", "close", queries, nil)
+	if err == nil {
+		// drain the buffer and close the body
+		_, _ = io.Copy(ioutil.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}
 }
 
 func (client *syncClient) syncHoleInterval(holeInterval Interval) error {
@@ -205,10 +220,14 @@ func (client *syncClient) syncHoleInterval(holeInterval Interval) error {
 	if err != nil {
 		return fmt.Errorf("sendHole failed, err: %s", err)
 	}
+
+	// drain the buffer and close the body
+	_, _ = io.Copy(ioutil.Discard, resp.Body)
+	_ = resp.Body.Close()
+
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	resp.Body.Close()
 
 	return nil
 }
@@ -221,12 +240,16 @@ func (client *syncClient) getServerChecksum(checksumInterval Interval) ([]byte, 
 	if err != nil {
 		return nil, fmt.Errorf("getChecksum failed, err: %s", err)
 	}
+
+	// drain the buffer and close the body
+	checksum, err := ioutil.ReadAll(resp.Body)
+	_ = resp.Body.Close()
+
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	defer resp.Body.Close()
 
-	return ioutil.ReadAll(resp.Body)
+	return checksum, err
 }
 
 func (client *syncClient) writeData(dataInterval Interval, data []byte) error {
@@ -237,10 +260,14 @@ func (client *syncClient) writeData(dataInterval Interval, data []byte) error {
 	if err != nil {
 		return fmt.Errorf("writeData failed, err: %s", err)
 	}
+
+	// drain the buffer and close the body
+	_, _ = io.Copy(ioutil.Discard, resp.Body)
+	_ = resp.Body.Close()
+
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
 	}
-	resp.Body.Close()
 
 	return nil
 }
--- a/sparse/file.go
+++ b/sparse/file.go
@@ -1,11 +1,13 @@
 package sparse
 
 import (
+	"context"
 	"crypto/sha512"
 	"fmt"
 	"io"
 	"os"
 	"syscall"
+	"time"
 	"unsafe"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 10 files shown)


sparse/file.go
```diff
@@ -192,46 +194,186 @@ func HashData(data []byte) ([]byte, error) {
...
+				errc <- err
+				return
+			}
+
+			intervals := make([]FileInterval, 0, len(exts))
+			isFinalInterval = len(exts) == 0
+			totalExtents += len(exts)
+
+			// special case the whole file is a hole
+			if totalExtents == 0 && fileInfo.Size() != 0 {
```
shuo-wu, line 241: Is the file size not empty caused by the hole info in the fiemap?

sparse/file.go
```diff
@@ -192,46 +194,186 @@ func HashData(data []byte) ([]byte, error) {
...
+					// add a hole between last data segment and end of file
+					if lastIntervalEnd < fileInfo.Size() {
+						hole := FileInterval{SparseHole, Interval{lastIntervalEnd, fileInfo.Size()}}
+						intervals = append(intervals, hole)
+					}
+				}
+			}
+
+			// transmit intervals
+			for _, interval := range intervals {
```
shuo-wu, line 269: Why does this function append data chunks or holes to the interval list first then consume them here, rather than directly push each interval to the buffered channel `out`?

sparse/client.go
```diff
@@ -104,53 +104,74 @@ func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64,
...
-			}
-			segmentStartTime := time.Now()
-			if segment.Kind == SparseHole {
-				if err = client.syncHoleInterval(segment.Interval); err != nil {
-					return fmt.Errorf("syncHoleInterval %s failed, err: %s", segment.Interval, err)
-				}
-			} else if segment.Kind == SparseData {
-				if err = client.syncDataInterval(file, segment.Interval); err != nil {
-					return fmt.Errorf("syncDataInterval %s failed, err: %s", segment.Interval, err)
+	mutex := &sync.Mutex{}
```
shuo-wu: It seems that this lock mainly protects variable `count`, which is also for benchmarking.

sparse/file.go
```diff
@@ -157,7 +157,7 @@ func alignmentShift(block []byte) int {
 }
 
 func ReadDataInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
-	data := make([]byte, dataInterval.Len())
+	data := AllocateAligned(int(dataInterval.Len()))
```
shuo-wu, line 160: I am not sure why you prefer to do alignment here instead of `file.ReadAt`. If the file processor is a buffered one, alignment here doesn't make sense.

sparse/test/fiemap_test.go
```diff
@@ -0,0 +1,337 @@
...
+	totalCount := 0
+	fiemap := NewFiemapFile(fileIo.GetFile())
+
+	// retrieve total extent count
+	//count, _, errno := fiemap.FiemapRegion(0, extStart, extLength)
+	//if errno != 0 { t.Fatal("failed to call fiemap.Fiemap(0)") }
+	//log.Infof("count retrieval done, time elapsed: %.2f second, extent number: %v", time.Now().Sub(startTime).Seconds(), count)
+
+	deltaTime := time.Now()
+	for {
```
shuo-wu: Can the test directly call the related function rather than having a duplicate logic here?

sparse/file.go
```diff
@@ -192,46 +194,186 @@ func HashData(data []byte) ([]byte, error) {
...
 	}
-	return GetFiemapRegionExts(file, Interval{0, fileInfo.Size()})
+
+	const MaxExtentsBuffer = 1024
+	const MaxInflightIntervals = MaxExtentsBuffer * 10
+	out := make(chan FileInterval, MaxInflightIntervals)
+	errc := make(chan error, 1)
+	go func() {
+		defer close(out)
+		defer close(errc)
```
shuo-wu, line 214: `close()` can happen before the receiver handles the signal

sparse/file.go
```diff
@@ -192,46 +194,186 @@ func HashData(data []byte) ([]byte, error) {
...
+			}
+		}
+	}()
+
+	return out, errc, err
 }
 
-func GetFiemapRegionExts(file FileIoProcessor, interval Interval) ([]Extent, error) {
-	if interval.End == 0 {
+func getFiemapExtentsIterative(file FileIoProcessor) ([]Extent, error) {
```
shuo-wu: What is this function for? unit test?

sparse/client.go
```diff
@@ -68,70 +88,79 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
...
-				return fmt.Errorf("syncHoleInterval %s failed, err: %s", holeInterval, err)
+	mutex := &sync.Mutex{}
+	processSegment := func(segment FileInterval) error {
+		segmentStartTime := time.Now()
+		if segment.Kind == SparseHole {
+			if err := client.syncHoleInterval(segment.Interval); err != nil {
+				return fmt.Errorf("syncHoleInterval %s failed, err: %s", segment.Interval, err)
+			}
+		} else if segment.Kind == SparseData {
+			if err := client.syncDataInterval(file, segment.Interval); err != nil {
```
shuo-wu, line 110: We may need to split a huge interval into smaller chunks during sending or calculating the checksum:
https://github.com/longhorn/longhorn/issues/1948#issuecomment-791641308

shuo-wu (changes requested):
1. There are some unnecessary comments in the unit tests. We can remove them later.
2. There are too many logs for the benchmarking. Besides, some of them are in `debug` level rather than `trace` level.

sparse/client.go
joshimoo: I plan on removing the benchmarking code in the final version since it's not needed same with the lock, I still need to do the incluster benchmark :)

sparse/file.go
joshimoo: Can you explain your concern?

sparse/test/fiemap_test.go
joshimoo: These tests aren't needed anymore, since the new fiemap code is covered by all sync tests.
This is just benchmarking code that can be removed, in the final version.

sparse/file.go
joshimoo: Not needed anymore, bechmarking code

sparse/client.go
joshimoo: The client.syncDataInterval already does chunking of the data interval (32 * 4K)
We might want to add additional concurrency for the individual chunks of a data interval later.
But even without that addition we already get 4x concurrency, since we process 4 intervals at a time.

sparse/file.go
joshimoo: makes the code nicer, since it separates the extent processing from the interval processing.
Think of this as: 
- map (extent -> interval) 
- reduce(interval)

sparse/file.go
joshimoo: If the filesize is 0, there is nothing to be done.
This is just an extra guard I kept, but it's not needed :)
```
		isFinalInterval := fileInfo.Size() == 0
		for !isFinalInterval {
```

sparse/file.go
shuo-wu: For example, if getting fiemap extents failed and the goroutine returns, is it possible that the error channel may get closed before the receiver handle the error signal (the receiver will check if the channel is closed first)? If yes, then the error will be lost and the receivers wrongly consider everything works fine.
Similar concern for the last signal of the interval channel.

sparse/file.go
shuo-wu: Oh, this is just a question. I am curious why fileInfo.Size() is not 0 when the whole file is a hole.

sparse/file.go
shuo-wu: I am thinking if this will degrade the performance a little bit since this is like a barrier in the med of a pipeline. 
For example, the below signal sending is suddenly blocked for a while, everything will get stuck. But actually, this goroutine can continue filling up the interval channel as long as the channel buffer is not full.

sparse/file.go
joshimoo: Thanks for clarifying, I understand the question now.
Let me give an example + explanation:

```shell
$ truncate --size=10K test.truncate
$ ls -lsah test.truncate 
0 -rw-rw-r-- 1 jmoody jmoody 10K May 26 15:00 test.truncate

$ fiemap test.truncate 
File test.truncate has 0 extents:
#	Logical          Physical         Length           Flags
```

From the above example we can see that a file that only consists of a hole, consumes 0 actual space, while the fileinfo.Size() would return 10K as that is specified size for that file.

Since the initial file gets created during the serverOpen call on the remote end, which does a truncate to the specified fileSize. In the 0 size case there is no need to do anything, since the remote file will already be created with 0 size.

sparse/file.go
joshimoo: Fair point, let me explain why I decide to do this here:
- allocating an aligned buffer is just as cheap as allocating len(unaligned buffer + 4K)
- this is since all this does is allocate a buffer + 4K then return a slice of that buffer where the base[0] points to an aligned memory address.

Now for the case where it really matters, directIO
- in the case where the randomly allocated buffer is unaligned:
-- we need to do 2 buffer allocations (2x the memory)
-- a full buffer copy after the read

Both of these operations are unnecessary, considering that we are most likely going to be using direct IO.
But even if not, we don't really incur much additional cost, but save costs for each ReadInterval call that is done with directIO (Data read, Checksum, etc)

------------------------------------------------------------
Summary:
In the case where we are using direction IO, we save an extra len(buf + 4K) allocation + memory copy.
In the case where we use buffered IO, we carry an extra (4K memory allocation) for the lifetime of that buffer.

What do you think? Any concerns?

sparse/file.go
shuo-wu: OK. `fileinfo.Size() ` returns apparent size rather than actual size. I was thinking it returned the actual size, lol.

sparse/file.go
shuo-wu: Thank you for the explanation. It's a really nice optimization!

sparse/file.go
joshimoo: This should not be an issue, let me try walking you through your mentioned case, on the example of the errc channel, same applies to the out channel.

```go
	out, errc, err := GetFileLayout(ctx, file)
        // err itself is prestart of the go routine only, so nothing to worry about, since we will exit

	// let's say the go routine has nothing todo (fileInfo.Size() == 0) case
	// so it returns immediately, and closes the errc, out channels that we assigned above
	// I am guessing this is your concerned case correct?

	const WorkerCount = 4
	errorChannels := []<-chan error{errc}
	for i := 0; i < WorkerCount; i++ {
		errorChannels = append(errorChannels, processFileIntervals(ctx, out, processSegment))
	}

	// since out is closed already as well, the individual processFileInterval go routines have nothing to do and exit as well
	// you can think through both cases, they will close later or are closed already here, this makes no difference in regard
	// to the correctness of the solution, so let's go with the they have closed their error channels already

	// the below select will exit once all error channels are closed, or a single
	// channel has run into an error, which will lead to the ctx being cancelled
	mergedErrc := mergeErrorChannels(ctx, errorChannels...)
	select {
	case err = <-mergedErrc:
		break
	}

	// above we have merged all the error channels and are waiting for a signal on the mergedErrc out channel.
        // let's have a look at how the mergeErrorChannels actuall work
        return err
```
-------------
```go
// mergeErrorChannels will merge all error channels into a single error out channel.
// the error out channel will be closed once the ctx is done or all error channels are closed
// if there is an error on one of the incoming channels the error will be relayed.
func mergeErrorChannels(ctx context.Context, channels ...<-chan error) <-chan error {
	var wg sync.WaitGroup
	wg.Add(len(channels))

	// here we create a buffered output channel, for each go routine one slot
	// for the worst case scenario where each go routines errors at the same time
	// this also prevents them from getting stuck trying to send an error
	out := make(chan error, len(channels))
	output := func(c <-chan error) {
		defer wg.Done()
		select {
		case err, ok := <-c:
			if ok {
				out <- err
			}
			return
		case <-ctx.Done():
			return
		}
	}

	// create a relaying go routine for each of the incoming error channels
	// that rely the error into the output channel and exit
	for _, c := range channels {
		go output(c)
	}

	// we wait till all go routines have exited, either because they had an error
	// the ctx was cancelled as a result of a single error
	// or all the incoming error channels where closed (this is the case that you are concerned about)
	// afterwards we close our error channel and exit this go routine.
	go func() {
		wg.Wait()
		close(out)
	}()
	return out
}
```

------------------

As you see there is should no be any issue.
I think your concern stems from the fact that you are concerned about reading from a closed channel:
- one will continuously read the default value from a closed channel, so it's fine reading from a closed channel
- one can use the two return form to determine whether a channel is closed `case err, open := <-c:`
- all producer go routines use a buffered errc[1] channel, so they don't block
- buffered channels are drained first in / first out (d1,d2, close, close, close, close, close, etc)

On any error the ctx will be cancelled so all go routines will terminate and each channel closed.
Once all go routines are closed, there will be no more references to the channels and they can be garbage collected even if they still have unprocessed data (early exit interval processing do to error)

-----------------

Let me know if that helps clarify the situation, if not we can do a peer review and discuss it while going over the code.

joshimoo:
Thanks for the good questions on this pr @shuo-wu 
The answers & explanations should lead to some good knowledge sharing for all of the team.

sparse/file.go
joshimoo: Another fair point, good that you asked. Let me elaborate on why I picked this design.

The base assumption / case here is that the rate of **inflow > outflow**, 
- This holds since the time requirement of processing a interval is much higher than the retrieval of the extents.
- Another thing to note is that the processing of intervals is io bound and not cpu bound.
- Which makes it a good candidate for concurrency to begin with.

If the rate of inflow < outflow then a trickle down pipeline (push individual intervals) might be better, but in that case one should also look into reducing the outflow (worker - 1) to further reduce the wait time.

------------------------------------------------------------------------------------------------

So why this design? (of course the analogy has some leaky abstractions :)
Here is an analogy, we have a 10L bucket, with 4 holes at the bottom. 
If we let water slowly drip into it (individual segments) then there will be a lot more contention at the beginning but the startup delay is reduced since each drip can go out of a single hole, while the 3 other holes wait for the next drip. 
If instead we fill up a 1L bottle first then dump all that water in at once, all the holes will have a continuous outflow (even if there is variance at the rate of outflow). This comes with a small startup delay (first 1024 extents) but in our case doesn't matter.

------------------------------------------------------------------------------------------------

Enough with the analogies, let's talk about the code:
- splits the concerns of map(extent -> interval) and reduce (drain interval)
- makes it easier to read the code, since special hole case / hole / data intervals all end up using the same code path.
- single dedicated place where we write into the channel, this means there is also only a single place where the routine can block. This also means that when the channel gets unblocked, there is only a single place for the scheduler to return and the next item will be immediately pushed back into the channel.
- this allows the cpu to cache the interval buffer, since there are no more writes to it.
- this separation makes it really easy to split/merge data intervals if we want to add later.
- we want to have a buffered channel, I used (10 * MAX_EXTENTS) so we have bounded memory consumption

---------------------------------------------------------------------------

These are some of the reasons for the above design. Using a trickle down style pipeline would be just as valid, but would make the code less easy to read, since there would be many places where you could end up blocking (buffer full) this makes it harder to benchmark/analyse a runtime profile.

Let me know what you think :)

sparse/file.go
shuo-wu: Got it. Reading a closed channel should not be an issue. I am sorry for misremembering this golang mechanism. Your explanation is pretty clear.

sparse/file.go
shuo-wu: Got it. We can handle the production and the consumption separately if the delay of the production can be ignored.
And the benefits look good. But I don't understand these 2 parts, can you explain more about them?
1. `there is only a single place for the scheduler to return and the next item will be immediately pushed back into the channel`
2. `this allows the cpu to cache the interval buffer, since there are no more writes to it.`

sparse/client.go
shuo-wu: OK. I forgot that part. Then the current optimization is enough now.

sparse/file.go
joshimoo: Not a problem, good that you asked. 
What is an issue is if you are reading from a nil channel since that will always block.
As well as sending to a closed/nil channel which will panic.

joshimoo:
I removed the benchmark code in the latest commit. I am leaving them in the pr so that if we need to benchmark again we can just cherry pick/revert that commit in the future :)

joshimoo:
@shuo-wu I marked some of the issues as unresolved, since it has some good questions + answers which can act as knowledge sharing for the rest of the team :)

sparse/file.go
joshimoo: Below are some good information on how the go scheduler works :)

Simple overview:
https://riteeksrivastava.medium.com/a-complete-journey-with-goroutines-8472630c7f5c#:~:text=The%20switch%20between%20Goroutines%20happen,if%20those%20operations%20would%20block.

Nice gotcha pre GO 1.14
http://www.sarathlakshman.com/2016/06/15/pitfall-of-golang-scheduler

Detailed Overview:
https://medium.com/@ankur_anand/illustrated-tales-of-go-runtime-scheduler-74809ef6d19b

New non cooperative preemptive scheduler GO 1.14:
https://github.com/golang/proposal/blob/master/design/24543-non-cooperative-preemption.md

-----------------------------------------------

This is more of witch craft, so you probably should it ignore it :)
There is also a bunch of leaky abstractions involved here :)

As long as the GO routine P, which would be doing the For Loop: Interval -> sent to channel
Can remain bound to Thread M and continue running for multiple executions instead of being preempted.
Since their are no writes to the underlying buffer, the CPU core that is executing thread M, should be able to load up the underlying array of the intervals slice into the cache instead of having to fetch each element from memory.

joshimoo:
@shuo-wu any outstanding feedback or can this be merged?

shuo-wu (dismissed):
In general LGTM. After removing the benchmarking code in `sfold.go`, it's good to go.

joshimoo:
@shuo-wu I squashed the commits and I still needed to add a lock for the UpdateFoldFileProgress (optimized this lock)
to ensure that there is only a single in flight file fold progress update. We can refactor this part later once we ensure that out of order progress updates are not a problem as well as look into switching the way progress updates are done.

The only thing that changes is the coalescing function.

shuo-wu (approved):

innobead:
@joshimoo Please help backport to 1.1.2.

PR #77, opened 2021-07-23
Implement SyncSnapshot which syncs the content of source snapshot to a file on remote host

PhanLe1010:
longhorn/longhorn#2403
longhorn/longhorn#2534
longhorn/longhorn#1815


sparse/test/client_test.go
```diff
@@ -797,3 +799,87 @@ func Benchmark_1G_CheckFiles(b *testing.B) {
...
+
+func TestSyncSmallSnapshot4MB(t *testing.T) {
+	// use the file at localPath to represent the source snapshot
+	localPath := tempFilePath("ssync-small-src-")
+	remotePath := tempFilePath("ssync-small-dst-")
+
+	filesCleanup(localPath, remotePath)
+	defer filesCleanup(localPath, remotePath)
+
+	size := 4194304 // 4194304 = 4MB
```
joshimoo: sometimes I do this in the inner scope
that way you don't need the comments
```
const MB = 1024 * 1024
size := 4 * MB
```

Alternative is which still is obvious that it's 4 MB
`size := 4 * 1024 * 1024`

The multiplications should be optmized out by the compiler :)

sparse/file.go
```diff
@@ -156,7 +156,7 @@ func alignmentShift(block []byte) int {
 	return int(uintptr(unsafe.Pointer(&block[0])) & uintptr(alignment-1))
 }
 
-func ReadDataInterval(file FileIoProcessor, dataInterval Interval) ([]byte, error) {
+func ReadDataInterval(file ReaderWriterAt, dataInterval Interval) ([]byte, error) {
```
shuo-wu: Rename the input to `rw` to make the naming consistent.

sparse/client.go
```diff
@@ -24,13 +24,18 @@ const (
 type syncClient struct {
 	remote   string
 	timeout  int
-	filePath string
-	fileSize int64
-	fileIo   FileIoProcessor
```
shuo-wu, line 29: One quick question, why not replacing `fileSize int64` & `fileIo FileIoProcessor` with `size int64` & `rw ReaderWriterAt`?

joshimoo (commented):
I will have another look later.

sparse/test/client_test.go
PhanLe1010: Thanks Joshua! Fixed

sparse/file.go
PhanLe1010: Thanks Shuo! Fixed

sparse/client.go
PhanLe1010: I don't see we use those fields any where in the code (except for the fileSize which is used once). So, I don't think we need to keep them inside the sync client struct.

shuo-wu: OK. I think we can retain `size` at least so that there is no need to add an extra input for some functions.

PhanLe1010: Can we also remove `size`? because:
1. We already pass down `filePath`, `FileIo` so `size` should be similar
2. I see that in other clients (`controller client`, `replica client`) we usually only keep the address of severs

shuo-wu (commented):
Notice that we can merge `syncSnapshotContent` into `syncFileContent`. But the main blocker is the function call `out, errc, err := GetFileLayout(ctx, file)` in `syncFileContent`. There are 2 ways to solve it.

1. we can try to get the channels outside of the function. Actually, the channel init can be done before the client init. 
    - For the snapshot sync, the caller SyncAgentServer can get the channels before calling the following function:
    ```
    func SyncContent(ctx context.Context, remote string, timeout int, directIO bool, size int64, rw ReaderWriterAt, outc <-chan FileInterval, errc <-chan error) error {
            // SyncClient doesn't care if the data is from a file or not.
	    client := newSyncClient(ctx, remote, timeout, directIO, rw, outc, errc)
	    defer client.closeServer()
    
	    syncStartTime := time.Now()
    
	    err := client.syncFileContent()
	    if err != nil {
		    log.Errorf("SyncContent failed: %s", err)
		    return err
	    }
    
	    log.Debugf("finished sync, size: %v elapsed: %.2fs",
		    size, time.Now().Sub(syncStartTime).Seconds())
	    return nil
    }
    ```

    - For file sync, function `SyncFile` can be refactored to call the above function as well:
    ```
    // SyncFile synchronizes local file to remote host
    func SyncFile(localPath string, remote string, timeout int, directIO bool) error {
	    // Validate files
	    fileInfo, err := os.Stat(localPath)
	    if err != nil {
		    log.Errorf("Failed to get size of source file: %s, err: %s", localPath, err)
		    return err
	    }
	    fileSize := fileInfo.Size()
	    if directIO && fileSize%Blocks != 0 {
		    return fmt.Errorf("Invalid directIO file block size: %v", fileSize)
	    }
	    log.Infof("source file size: %d, setting up directIO: %v", fileSize, directIO)
    
	    // Prepare ReaderWriterAt
	    var fileIo FileIoProcessor
	    if directIO {
		    fileIo, err = NewDirectFileIoProcessor(localPath, os.O_RDONLY, 0)
	    } else {
		    fileIo, err = NewBufferedFileIoProcessor(localPath, os.O_RDONLY, 0)
	    }
	    if err != nil {
		    log.Error("Failed to open local source file:", localPath)
		    return err
	    }
	    defer fileIo.Close()
    
	    // Prepare channels
	    ctx, cancel := context.WithCancel(context.Background())
	    defer cancel()
	    outc, errc, err := GetFileLayout(ctx, fileIo)
	    if err != nil {
		    return fmt.Errorf("failed to retrieve file layout for file %v err: %v", fileIo.Name(), err)
	    }
    
	    if err := SyncContent(ctx, remote, timeout, directIO, fileSize, fileIo, outc, errc); err != nil {
		    log.Errorf("syncFileContent failed: %s", err)
		    return err
	    }
    
	    return err
    }
    ```
    - Then, `client.syncFileContent` is responsible for getting intervals from channels and sending them to the remote only. It doesn't need to take care of the channel init:
    ```
    func (client *syncClient) syncFileContent(outc <-chan FileInterval, errc <-chan error) error {
	    if err := client.openServer(client.directIO); err != nil {
		    return fmt.Errorf("openServer failed, err: %v", err)
	    }
    
	    syncStartTime := time.Now()
    
	    processSegment := func(segment FileInterval) error {
		    return client.processSegment(rw, segment)
	    }
    
            const WorkerCount = 4
            ......
    }
    ```


2. we can define an interface for `ReaderWriterAt` so that the sync client can blindly get channels with that interface. 
    - Both the FileIO and the replica need to implement the interface. e.g.:
    ```
    func (file *BufferedFileIoProcessor) GetFileLayout(ctx context.Context) (<-chan FileInterval, <-chan error, error) {
	    return GetFileLayout(ctx, file)
    }
    
    func (file *DirectFileIoProcessor) GetFileLayout(ctx context.Context) (<-chan FileInterval, <-chan error, error) {
	    return GetFileLayout(ctx, file)
    }
    ```
    - Then `FileIoProcessor` in `syncClient` can be replaced by `ReaderWriterAt`


Notice that the above functions are not valid codes. If they make sense to you, you can refer to them then do refactoring.

sparse/client.go
joshimoo: > 2. I see that in other clients (`controller client`, `replica client`) we usually only keep the address of severs

GRPC client usage is changing in https://github.com/longhorn/longhorn-engine/pull/644
I still retain the serviceURL's for log printing.

PhanLe1010: Thank you all. I added the fields

PhanLe1010:
Thanks @shuo-wu !
I have implemented the 2nd method

sparse/client.go
```diff
@@ -74,65 +82,84 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
...
+		fileIo.Name(), fileSize, time.Now().Sub(syncStartTime).Seconds())
 
-	return err
+	return nil
 }
 
-func (client *syncClient) syncFileContent(file FileIoProcessor, fileSize int64, directIO bool) error {
-	if err := client.openServer(directIO); err != nil {
-		return fmt.Errorf("openServer failed, err: %v", err)
+func SyncSnapshot(snapshotName string, rw ReaderWriterAt, snapshotSize int64, remote string, timeout int) error {
```
shuo-wu: Rename this to `SyncContent(rw ReaderWriterAt, size int64, remote string, timeout int) error` so that we can remove the concept `snapshot` from this library. Besides, `SyncFile` can call the renamed function as well.

sparse/client.go
```diff
@@ -24,13 +24,21 @@ const (
 type syncClient struct {
 	remote   string
 	timeout  int
-	filePath string
-	fileSize int64
-	fileIo   FileIoProcessor
+	rwName string
+	rwSize int64
```
shuo-wu: `sourceName` `size`

shuo-wu (changes requested):

sparse/client.go
PhanLe1010: Thank you. Implemented

sparse/client.go
PhanLe1010: Thank you. Changed

sparse/client.go
```diff
@@ -74,65 +88,74 @@ func SyncFile(localPath string, remote string, timeout int, directIO bool) error
 	}
 	defer fileIo.Close()
 
-	client := newSyncClient(remote, timeout, localPath, fileSize, fileIo)
+	return SyncContent(fileIo.Name(), SourceTypeFile, fileIo, fileSize, remote, timeout, directIO)
+}
+
+func SyncContent(sourceName string, sourceType SourceType, rw ReaderWriterAt, size int64, remote string, timeout int, directIO bool) (err error) {
+	defer func() {
+		err = errorWrapf(err, "failed to SyncContent for source: %v, type: %v", sourceName, sourceType)
```
shuo-wu: You can use `errors.Wrapf()` instead

shuo-wu (changes requested):

sparse/client.go
PhanLe1010: I was hesitated to use `errors.Wrapf()` since we would have to add go module `github.com/pkg/errors` just for the function. But using standard function may be good in the long run so I changed it as your suggestion. Thanks

shuo-wu (dismissed):
LGTM. @joshimoo Can you take a final look?

joshimoo (approved):
LGTM, greate job @PhanLe1010  and thank you @shuo-wu for driving the review 👍🏻

PR #80, opened 2022-03-16
Snapshot prune

shuo-wu:
longhorn/longhorn#3613

### DIFF
--- a/sparse/sfold.go
+++ b/sparse/sfold.go
@@ -54,35 +54,15 @@ func FoldFile(childFileName, parentFileName string, ops FileHandlingOperations) 
 }
 
 func coalesce(parentFileIo, childFileIo FileIoProcessor, fileSize int64, ops FileHandlingOperations) (err error) {
-	var progress uint32
+	progress := new(uint32)
 	progressMutex := &sync.Mutex{}
-
-	// progress updates can only be monotonically increasing, once that's replaced we can use an atomic cmp&swp around progress
-	// that doesn't require locking and we also limit the update function calls to p=[0,100] since there
-	// is no point in calling updates for the same p value multiple times
-	updateProgress := func(newProgress uint32, done bool, err error) {
-		forcedUpdate := done || err != nil
-		if !forcedUpdate && newProgress <= atomic.LoadUint32(&progress) {
-			return
-		}
-
-		// this lock ensures that there is only a single in flight UpdateFoldFileProgress operation
-		// with multiple in flight operations, the lock on the receiver side would be non deterministically
-		// acquired which would potentially lead to non-monotonic progress updates.
-		progressMutex.Lock()
-		defer progressMutex.Unlock()
-		if forcedUpdate || newProgress > atomic.LoadUint32(&progress) {
-			ops.UpdateFileHandlingProgress(int(newProgress), done, err)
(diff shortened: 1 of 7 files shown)


sparse/sfold.go
```diff
@@ -63,36 +53,16 @@ func FoldFile(childFileName, parentFileName string, ops FoldFileOperations) erro
 	return coalesce(parentFileIo, childFileIo, childFInfo.Size(), ops)
 }
 
-func coalesce(parentFileIo, childFileIo FileIoProcessor, fileSize int64, ops FoldFileOperations) (err error) {
-	var progress uint32
+func coalesce(parentFileIo, childFileIo FileIoProcessor, fileSize int64, ops FileHandlingOperations) (err error) {
+	progress := new(uint32)
 	progressMutex := &sync.Mutex{}
 
-	// progress updates can only be monotonically increasing, once that's replaced we can use an atomic cmp&swp around progress
```
shuo-wu, line 70: Actually, I am not sure if we could replace CAS with the locking. Because CAS could handle variable `progress` update only but we need to make sure the variable update **and the interface call `ops.UpdateFoldFileProgress()`** atomic. 

Here is the replacement I tried but I don't think it would work:
```
func updateProgressCAS(currentProgressAddr *uint32, newProgress uint32, done bool, err error, progressMutex *sync.Mutex, ops FileHandlingOperations) {
	forcedUpdate := done || err != nil
	if forcedUpdate {
		ops.UpdateFileHandlingProgress(int(newProgress), done, err)
		atomic.StoreUint32(currentProgressAddr, newProgress)
		return
	}

	for currentProgress := atomic.LoadUint32(currentProgressAddr); currentProgress < newProgress; currentProgress = atomic.LoadUint32(currentProgressAddr) {
		if atomic.CompareAndSwapUint32(currentProgressAddr, currentProgress, newProgress) {
			ops.UpdateFileHandlingProgress(int(newProgress), done, err)
			return
		}
	}
}
```

innobead (approved):
LGTM

sparse/sfold.go
innobead: atomic-related functions only apply to the target pointer so outside of this scope, it can't control in the same atomic operation. As you said, the example here should not work.

shuo-wu: Then can we remove this comment?
cc @joshimoo

joshimoo: I don't really remember what I was thinking back then :)

But I think it was that I thought it might be better to replace the update progress api to send
currentByte, totalByte and have the receiver calculate the progress.

This would also allow the receiver to provide summary statistics and calculate expected runtime based on the previously observed changes.

My quick thought is that this should allow no locking on the receiver end, since the receiver can just ignore update calls with current < highest observed (similiar to CAS), since progress cannot decrease. 

But I don't really remember what I was thinking, @shuo-wu thanks for playing with this and trying out your change above :)

No further action needed here at the moment, feel free to remove the comments.

shuo-wu: Unlike file syncing, there is no receiver or sender during coalescing/pruning.
Or do you mean that using a separate goroutine with a channel as the receiver? The channel is actually using locking to protect the signal transmission.

PR #81, opened 2022-09-08
Sparse file support operation unmap

shuo-wu:
longhorn/longhorn#836

### DIFF
--- a/sparse/file.go
+++ b/sparse/file.go
@@ -17,8 +17,10 @@ type FileIoProcessor interface {
 	// File I/O methods for direct or bufferend I/O
 	ReadAt(data []byte, offset int64) (int, error)
 	WriteAt(data []byte, offset int64) (int, error)
+	UnmapAt(length uint32, offset int64) (int, error)
 	GetFile() *os.File
-	GetDataLayout (ctx context.Context) (<-chan FileInterval, <-chan error, error)
+	GetFieMap() *FiemapFile
+	GetDataLayout(ctx context.Context) (<-chan FileInterval, <-chan error, error)
 	Close() error
 	Sync() error
 	Truncate(size int64) error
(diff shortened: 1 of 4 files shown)


sparse/file.go
```diff
@@ -42,18 +45,32 @@ func NewBufferedFileIoProcessor(name string, flag int, perm os.FileMode, isCreat
...
+func (file *BufferedFileIoProcessor) UnmapAt(length uint32, offset int64) (int, error) {
+	return int(length), file.fieMap.PunchHole(offset, int64(length))
 }
 
 func (file *BufferedFileIoProcessor) GetFile() *os.File {
 	return file.File
 }
 
-func (file *BufferedFileIoProcessor) GetDataLayout (ctx context.Context) (<-chan FileInterval, <-chan error, error) {
+func (file *BufferedFileIoProcessor) GetFieMap() FiemapFile {
```
innobead: Suggest returning the reference `*FiemapFile` instead of creating another structure.

Also, it seems we can refactor `file.File` to private `file.file` in BufferedFileIoProcessor and DirectFileIoProcessor, because it seems all of them just be exported by `GetFile` method.

innobead (changes requested):
LGTM, just a minro request.

joshimoo (approved):
LGTM, not sure if we need the extra wrapper for FiemapFile since all the code seems to act on the os.File Fd

shuo-wu:
Does the extra wrapper mean the IO Processors?

innobead (approved):
LGTM.

innobead:
> LGTM, not sure if we need the extra wrapper for FiemapFile since all the code seems to act on the os.File Fd

We still need another wrapper for the below here.
```golang
func (file *BufferedFileIoProcessor) UnmapAt(length uint32, offset int64) (int, error) {
	return int(length), file.PunchHole(offset, int64(length))
}
```

cc @shuo-wu

PR #82, opened 2022-11-28
Rebuilding speedup: check change time and checksum before syncing disk files

derekbit:
[Longhorn 4783](https://github.com/longhorn/longhorn/issues/4783)
    

derekbit:
Runing e2e. Will rebase later.

sparse/rest/handlers.go
```diff
@@ -111,37 +119,62 @@ func (server *SyncServer) doOpen(request *http.Request) error {
 	return nil
 }
 
+func (server *SyncServer) getFileIo(directIO bool) (sparse.FileIoProcessor, error) {
```
innobead: This is for creating a new instance, so it can be named `newXXXX` instead

sparse/rest/handlers.go
```diff
@@ -111,37 +119,62 @@ func (server *SyncServer) doOpen(request *http.Request) error {
...
+			info := &types.SnapshotHashInfo{
+				Method:     checksumMethod,
+				Checksum:   checksum,
+				ChangeTime: changeTime,
+			}
+
+			err = util.SetSnapshotHashInfoToChecksumFile(server.filePath+".checksum", info)
+		}
+
+		log.WithError(err).Warnf("Failed ot set snapshot hash info to checksum file %v", server.filePath)
```
innobead: Should only log error when err is not nil.

sparse/rest/handlers.go
```diff
@@ -189,28 +222,84 @@ func (server *SyncServer) doGetChecksum(writer http.ResponseWriter, request *htt
...
 	}
 
+	writer.Header().Set("Content-Type", "application/json")
+	fmt.Fprint(writer, string(metadata))
+
 	return nil
 }
+
+func (server *SyncServer) getRecordedMetadataFromChecksumFile() ([]byte, error) {
+	f, err := os.Open(server.filePath + ".checksum")
```
innobead: `.checksum` shows twice, so suggest having this as constant.

sparse/rest/handlers.go
```diff
@@ -111,37 +119,62 @@ func (server *SyncServer) doOpen(request *http.Request) error {
...
+	if checksumMethod != "" && checksum != "" {
+		changeTime, err := util.GetFileChangeTime(server.filePath)
+		if err == nil {
+			info := &types.SnapshotHashInfo{
+				Method:     checksumMethod,
+				Checksum:   checksum,
+				ChangeTime: changeTime,
+			}
+
+			err = util.SetSnapshotHashInfoToChecksumFile(server.filePath+".checksum", info)
```
innobead, line 150: So this just means after sync, save the checksum and ctime from the sender?

sparse/client.go
```diff
@@ -52,80 +61,113 @@ func newHTTPClient() *http.Client {
...
 		return err
 	}
 	defer fileIo.Close()
 
-	return SyncContent(fileIo.Name(), fileIo, fileSize, remote, timeout, directIO)
+	return SyncContent(fileIo.Name(), fileIo, fileSize, remote, timeout, directIO, fastSync)
 }
 
-func SyncContent(sourceName string, rw ReaderWriterAt, size int64, remote string, timeout int, directIO bool) (err error) {
+func getFileIo(localPath string, directIO bool) (FileIoProcessor, error) {
```
innobead: Rename to `newFileIoProcessor`

sparse/client.go
```diff
@@ -138,14 +180,80 @@ func (client *syncClient) syncContent() error {
...
+
+	var info types.SnapshotHashInfo
+	if err := json.Unmarshal(metadata, &info); err != nil {
+		return "", errors.Wrap(err, "failed to unmarshal hash info")
+	}
+
+	return info.Checksum, nil
+}
+
+func getLocalDiskFileChangeTimeAndChecksum(sourceName string) (string, string, string, error) {
```
innobead: nit: when having more than 2 return vars, I would suggest having named return vars to make the protocol clear.

sparse/client.go
```diff
@@ -206,22 +314,25 @@ func (client *syncClient) openServer() error {
...
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("resp.StatusCode(%d) != http.StatusOK", resp.StatusCode)
+		return fmt.Errorf("failed to open server: status code=%v (%v)",
+			resp.StatusCode, http.StatusText(resp.StatusCode))
 	}
 
 	return nil
 }
 
 func (client *syncClient) closeServer() {
```
innobead: you renamed `openServer` to `open`, so it's better to have a consistent change here like `closeServer` to `close`.

util/hash.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/longhorn/sparse-tools/types"
+)
+
+const (
+	tmpFileSuffix = ".tmp"
+)
+
+func SetSnapshotHashInfoToChecksumFile(checksumFileName string, info *types.SnapshotHashInfo) error {
```
innobead, line 17: nit: add unit test.

innobead (changes requested):
LGTM, but some minor comments needed to resolve.

sparse/client.go
innobead: nit: I would suggest having an interface for `syncclient` to abstract the functions to map to the exposed functions of the remote sync server like `open, close, sendHole, writeData, getChecksum, getRecordedMetadata`, and then `SyncClient` can implement the interface, so we will have a good understanding for the protocol in between and good for the future unit testing. 

```go
	router.HandleFunc("/v1-ssync/open", server.open).Methods("GET")
	router.HandleFunc("/v1-ssync/close", server.close).Methods("POST")
	router.HandleFunc("/v1-ssync/sendHole", server.sendHole).Methods("POST")
	router.HandleFunc("/v1-ssync/writeData", server.writeData).Methods("POST")
	router.HandleFunc("/v1-ssync/getChecksum", server.getChecksum).Methods("GET")
	router.HandleFunc("/v1-ssync/getRecordedMetadata", server.getRecordedMetadata).Methods("GET")

```

sparse/rest/handlers.go
derekbit: Yes. In the case, the disk file might be missing or mismatched with the one in the sender side. The checksum file will be created by the information from sender. Then, the rehashing is not necessary.

sparse/rest/handlers.go
derekbit: Updated. Thank you.

sparse/rest/handlers.go
derekbit: Updated. Thank you.

sparse/client.go
derekbit: Updated. Thank you.

sparse/rest/handlers.go
derekbit: Updated. Thank you.

sparse/client.go
```diff
@@ -52,51 +61,84 @@ func newHTTPClient() *http.Client {
...
 	}
-	log.Infof("source file size: %d, setting up directIO: %v", fileSize, directIO)
 
-	var fileIo FileIoProcessor
-	if directIO {
-		fileIo, err = NewDirectFileIoProcessor(localPath, os.O_RDONLY, 0)
-	} else {
-		fileIo, err = NewBufferedFileIoProcessor(localPath, os.O_RDONLY, 0)
-	}
+	log.Infof("Syncing file %v: size %v, directIO %v, fastSync %v", localPath, fileSize, directIO, fastSync)
```
innobead: You can also add `remote` to the log

innobead (dismissed):
LGTM

util/hash.go
derekbit: Addressed. Thank you.

sparse/client.go
derekbit: Updated, thank you.

util/hash_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func randomID(randomIDLenth int) string {
+	id := uuid.New().String()
+	return id[:randomIDLenth]
+}
+
+func TestSetSnapshotHashInfoToChecksumFile(t *testing.T) {
+	assert := require.New(t)
+
+	checksumFileName := randomID(8) + types.DiskChecksumSuffix
```
innobead, line 20: nit: it's better to delete the file in teardown via defer.

sparse/rest/server.go
```diff
@@ -9,6 +9,15 @@ import (
 	"github.com/longhorn/sparse-tools/sparse"
 )
 
+type DataSyncServer interface {
```
innobead, line 12: This is good! 

However, because this interface is only used internally (rest pkg), so suggest renaming all methods to lower camel case.

sparse/client.go
```diff
@@ -9,26 +9,44 @@ import (
...
 
 const (
-	httpClientTimeout = 5
-	numBlocksInBatch  = 32
+	httpClientTimeout      = 5
+	numBlocksInBatch       = 32
+	defaultSyncWorkerCount = 4
 )
 
+type DataSyncClient interface {
```
innobead, line 30: The same, because it's used internally, so renaming methods to lower camel case intead.

innobead (changes requested):

util/hash_test.go
derekbit: Addressed. Thank you.

sparse/client.go
derekbit: Addressed. Thank you.

innobead (approved):

PR #83, opened 2022-12-19
Speed up rebuilding a newly created replica on a node

derekbit:
Speed up rebuilding a newly created replica on a node rather than using a failed replica.
1. Avoid unnecessary local data checksum calculation
2. Increase the chunk size to 2 MiB

Longhorn/longhorn#4092

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -22,9 +22,10 @@ import (
 )
 
 const (
-	httpClientTimeout      = 5
-	numBlocksInBatch       = 32
+	httpClientTimeout = 5
+
 	defaultSyncWorkerCount = 4
+	defaultSyncBatchSize   = 512 * Blocks
 )
 
 type DataSyncClient interface {
@@ -47,6 +48,10 @@ type syncClient struct {
 	recordedChangeTime     string
 	recordedChecksumMethod string
 	recordedChecksum       string
+
+	fileAlreadyExistsOnServer bool
+	syncBatchSize             int64
+	numSyncWorkers            int
 
 	httpClient *http.Client
 }
@@ -71,7 +76,7 @@ func newHTTPClient() *http.Client {
 }
 
 func newSyncClient(remote string, timeout int, sourceName string, size int64, rw ReaderWriterAt, directIO bool,
-	recordedChangeTime, recordedChecksumMethod, recordedChecksum string) *syncClient {
+	recordedChangeTime, recordedChecksumMethod, recordedChecksum string, syncBatchSize int64, numSyncWorkers int) *syncClient {
 	return &syncClient{
 		remote:     remote,
 		timeout:    timeout,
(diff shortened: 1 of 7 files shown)


sparse/rest/handlers.go
```diff
@@ -103,6 +105,11 @@ func (server *SyncServer) doOpen(request *http.Request) error {
 	}
 	log.Infof("Receiving %v: size %v, directIO %v", server.filePath, interval.End, directIO)
 
+	if _, err := os.Stat(server.filePath); err != nil && errors.Is(err, os.ErrNotExist) {
+		log.Infof("file %v does not exist", server.filePath)
+		server.fileAlreadyExists = false
+	}
```
innobead: Suggest moving this to the constructor method `Server` of SyncServer, so you don't need to calculate this here. For thread-safe purpose, even we know internally it will be only called once.

sparse/client.go
```diff
@@ -327,15 +342,19 @@ func (client *syncClient) open() error {
 	}
 
 	// drain the buffer and close the body
-	_, _ = io.Copy(ioutil.Discard, resp.Body)
+	body, err := ioutil.ReadAll(resp.Body)
```
innobead, line 345: Should use `io.ReadAll`, because iotuil has been deprecated.

sparse/client.go
```diff
@@ -371,25 +390,34 @@ func (client *syncClient) syncHoleInterval(holeInterval Interval) error {
...
+			batchInterval, resp.StatusCode, http.StatusText(resp.StatusCode))
+	}
+
+	if err != nil {
+		return nil, err
 	}
 
-	return checksum, err
+	var checksum string
+	if err := json.Unmarshal(body, &checksum); err != nil {
```
innobead, line 416: ~Does this mean the previous `getServerChecksum` always not working correctly? due to a wrong decode.~ the decoding part should be moved to the getServerChecksum of the client.

innobead (changes requested):

innobead (commented):
this repo is also used by backing image manager and volume clone. See if need to update. 

cc @longhorn/qa

sparse/rest/handlers.go
derekbit: Updated. Thank you.

sparse/client.go
derekbit: Updated. Thank you.

sparse/client.go
```diff
@@ -371,25 +390,34 @@ func (client *syncClient) syncHoleInterval(holeInterval Interval) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var checksum []byte
+	if err := json.Unmarshal(body, &checksum); err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal checksum for interval %+v from server", batchInterval)
+	}
+
+	return checksum, nil
```
innobead, line 420: `body` is []byte already, so why need to decode here again? decoding is much for struct type. It seems the server encoding the byte array as well, so it's fine. However, this looks quite weird to me. :thinking:

innobead (approved):

PR #84, opened 2022-12-20
Make HTTP client timeout configurable

derekbit:
Longhorn/longhorn#5110

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	httpClientTimeout = 5
+	fileSyncOpenTimeout = 120
 
 	defaultSyncWorkerCount = 4
 	defaultSyncBatchSize   = 512 * Blocks
@@ -39,7 +39,6 @@ type DataSyncClient interface {
 
 type syncClient struct {
 	remote     string
-	timeout    int
 	sourceName string
 	size       int64
 	rw         ReaderWriterAt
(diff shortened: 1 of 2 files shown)


sparse/client.go
```diff
@@ -96,7 +96,7 @@ func newSyncClient(remote string, timeout int, sourceName string, size int64, rw
 }
 
 // SyncFile synchronizes local file to remote host
-func SyncFile(localPath string, remote string, timeout int, directIO, fastSync bool) error {
+func SyncFile(localPath string, remote string, httpClientTimeout int, directIO, fastSync bool) error {
```
innobead, line 99: This change will impact backing image manager, because it will use this exported function.

sparse/client.go
```diff
@@ -286,7 +285,7 @@ func (client *syncClient) processSegment(segment FileInterval) error {
 func (client *syncClient) sendHTTPRequest(method string, action string, queries map[string]string, data []byte) (*http.Response, error) {
 	httpClient := client.httpClient
 	if httpClient == nil {
-		httpClient = newHTTPClient()
+		httpClient = newHTTPClient(client.httpClientTimeout)
```
innobead, line 288: Why we need to reinitialize this here? `client.httpClient` should be already initialized when creating at `newSyncClient`. (no need to change this now to avoid any side effect)

innobead (approved):
LGTM

sparse/client.go
derekbit: Yeah, the backing image manager needs a update.

PR #90, opened 2023-01-10
sync file: buffering read disk files for replica rebuild

derekbit:
Longhorn/longhorn#5130

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -110,7 +110,7 @@ func SyncFile(localPath string, remote string, httpClientTimeout int, directIO, 
 
 	log.Infof("Syncing file %v to %v: size %v, directIO %v, fastSync %v", localPath, remote, fileSize, directIO, fastSync)
 
-	fileIo, err := newFileIoProcessor(localPath, directIO)
+	fileIo, err := NewBufferedFileIoProcessor(localPath, os.O_RDONLY, 0)
 	if err != nil {
 		log.WithError(err).Errorf("Failed to open local source file %v", localPath)
 		return err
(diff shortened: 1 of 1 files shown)


sparse/client.go
```diff
@@ -110,7 +110,7 @@ func SyncFile(localPath string, remote string, httpClientTimeout int, directIO,
 
 	log.Infof("Syncing file %v to %v: size %v, directIO %v, fastSync %v", localPath, remote, fileSize, directIO, fastSync)
 
-	fileIo, err := newFileIoProcessor(localPath, directIO)
+	fileIo, err := NewBufferedFileIoProcessor(localPath, os.O_RDONLY, 0)
```
joshimoo, line 113: NIT: why can we remove the directIO path here?

derekbit: For the server side (healthy replica), the direct IO is not really necessay because it is opened with read-only mode.
Using buffering read has some advantages
1. Leverage the filesystem's read perfetching.
2. If the replica rebuilding somehow failed, the new replica rebuilding can be accelerated, because the previous data are already in memory.

For the client side, the direct IO is still necessary.

derekbit: For the client side (healthy replica), the direct IO is not really necessay because it is opened with read-only mode.
Using buffering read has some advantages
- Leverage the filesystem's read perfetching.
- If the replica rebuilding somehow failed, the new replica rebuilding can be accelerated, because the previous data are already in memory.

For the server side (failed replica to be rebuilt), the direct IO is still necessary.

innobead (approved):
LGTM

PR #91, opened 2023-02-16
Revert "sync file: buffering read disk files for replica rebuild"

derekbit:
This reverts commit 727da1d6548d7d79a9095710cc56c77194ad3b08.

The instance-manager-r's memory usage increases due to the cache memory. 
After the rebuilding, the memory usage of the instance-manager-r increases. This is not harmful. However, to avoid uncertainties caused by the checks of pods' memory usages, we revert the commit.
    
Longhorn/longhorn#5130


innobead (dismissed):

derekbit:
Updated the comment. Thank you.

innobead (approved):

PR #94, opened 2023-03-27
chore: the io/ioutil package is deprecated from Go 1.16

jenting:
### Propose Changes
- The `io/ioutil` package is deprecated.
- Update the base Go version from 1.13 to 1.17.


innobead (changes requested):
Let's update go version to 1.17 to align with all components and libs.

innobead (approved):

innobead:
@jenting Please help update the dependencies of the dependent components in the master and 1.4.x branches. They should be longhorn-engine and backing-image-manager. 

cc @longhorn/dev-data-plane

jenting:
> @jenting Please help update the dependencies of the dependent components in the master and 1.4.x branches. They should be longhorn-engine and backing-image-manager.
> 
> cc @longhorn/dev-data-plane

@innobead There is the dependabot configuration triggered every week in longhorn-engine and backing-image-manager. Shouldn't we rely on the bot to update the dependency?

innobead:
@jenting Sounds good, but unfortunately, this is quite different from external dependencies, since the change of our shared libraries would immediately impact the functional integration. However, the idea is good and reasonable for sure.

PR #96, opened 2024-01-16
fix: return the whole data buf

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6899

When sending the batch with the interval, we allocate the `buf` with the batch size (interval size) and read data from the volume into it.
Then we send the `buf` as the data with the interval information to the target volume to write the data in that interval
We should not cut the data with the count we actually read from volume because that would be incorrect in the below example
```
Assuming we have a Volume and a snapshot chain
0 nil
1 BackingImage [0:2MB]
2 Snapshot1 [0:16MB]
 
We write some data to the Volume [3MB:4MB]

When we are exporting the Volume with block size 2MB, 
the preload index map will indicate that for [2MB:4MB] data
[2MB:3MB] should be read BackingImage
[3MB:4MB] should be read from snap1

Then since [2MB:3MB] is out of the bound of BackingImage,
we can't read the volume and the read count will be 0
[3MB:4MB] has the data and the read count would be 1MB

So the `buf` we read from the volume will be [0,0,0,.....0, x, x,x, ..., x]
If we cut the return data with the count, then we only get 1MB of [0,0,0,...,0] 
```

#### Special notes for your reviewer:
This function is a util function used in all sync content related feature.
I think if we test volume clone feature with volume expansion we might encounter the same issue.

#### Additional documentation or context

cc @shuo-wu @PhanLe1010

### DIFF
--- a/sparse/file.go
+++ b/sparse/file.go
@@ -196,7 +196,7 @@ func ReadDataInterval(rw ReaderWriterAt, dataInterval Interval) ([]byte, error) 
 			return nil, errors.Wrapf(err, "failed to read interval %+v", dataInterval)
 		}
 	}
-	return data[:n], nil
+	return data, nil
 }
 
 func WriteDataInterval(file FileIoProcessor, dataInterval Interval, data []byte) error {


ChanYiLin:
## Test is now running...
core e2e test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6002/console

shuo-wu (approved):

ChanYiLin:
## Test is now PASSED

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6002/

PhanLe1010:
I would like to review this PR but currently working on a support case. Will review it soon today.

PhanLe1010 (approved):
LGTM.

PhanLe1010:
cc @james-munson Do you have experience with the convention about the behavior when reading a block device outside of its size? Is it common to return n bytes of 0s and nil error?

ChanYiLin:
I am thinking if we are now returning the correct count even the BackingImage is smaller than Volume
We don't need this PR anymore

Furthermore, since the ReadAt semantic is aligned to the golang io ReadAt, we need to consider err==EOF, and only return valid data
```
n, err := rw.ReadAt(data, dataInterval.Begin)
```
we should keep the original code to cut the data with valid count.

WDYT? cc @shuo-wu @derekbit @innobead

innobead (commented):
yeah agreed, as you alrady return the right count in the engine PR.

PR #102, opened 2024-04-30
ci: change to use GitHub actions

FrankYang0529:
https://github.com/longhorn/longhorn/issues/8039

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+name: build
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+jobs:
+  build:
+    name: Build binaries
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    # Build binaries
+    - name: Run ci
+      run: make ci
+
+    - uses: codecov/codecov-action@v4
+      with:
+        files: ./coverage.out
+        flags: unittests
+        token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
+name: build
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+jobs:
+  build:
+    name: Build binaries
+    runs-on: ubuntu-latest
```
derekbit, line 10: Just curious, do we have other options?

FrankYang0529: Here, https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

FrankYang0529:
Hi @derekbit, it looks like test failed because I added error check to avoid lint issue. Can we assert `err` is `nil` here?

https://github.com/longhorn/sparse-tools/blob/813ac9e807f839eabd9f3583f843fcfea1269f0d/sparse/test/client_test.go#L255-L256

FrankYang0529:
> Hi @derekbit, it looks like test failed because I added error check to avoid lint issue. Can we assert `err` is `nil` here?
> 
> https://github.com/longhorn/sparse-tools/blob/813ac9e807f839eabd9f3583f843fcfea1269f0d/sparse/test/client_test.go#L255-L256

It looks like server returns `http: Server closed` when server is closed. I update the test case.

FrankYang0529:
Hi @derekbit, I tried to remove error check, but there still has data race condition. May you take a look when you have time? Thank you.

https://github.com/longhorn/sparse-tools/actions/runs/8888699426/job/24405917283?pr=102

shuo-wu:
I am fine with merging this then using another ticket to track the data race

PR #104, opened 2024-07-01
feat(auto-balance): disk in pressure

c3y1huang:

Issue longhorn/longhorn#4105

Implement a new function to sync data contents by local files.

#### Special notes for your reviewer:

`None`

#### Additional documentation or context

https://github.com/longhorn/longhorn-engine/pull/1003#pullrequestreview-2150095135

### DIFF
--- a/sparse/local.go
+++ b/sparse/local.go
@@ -41,9 +41,8 @@ func syncLocalContent(sourceFilePath, targetFilePath string, fileSize int64) (er
 	}()
 
 	syncBatchSize := defaultSyncBatchSize
-	numSyncWorkers := defaultSyncWorkerCount
 
-	local, err := newSyncLocal(sourceFilePath, targetFilePath, fileSize, syncBatchSize, numSyncWorkers)
+	local, err := newSyncLocal(sourceFilePath, targetFilePath, fileSize, syncBatchSize)
 	if err != nil {
 		return err
 	}
@@ -73,11 +72,10 @@ type syncLocal struct {
 	sourceFileIo *DirectFileIoProcessor
 	targetFileIo *DirectFileIoProcessor
 
-	numSyncWorkers int
-	syncBatchSize  int64
+	syncBatchSize int64
 }
 
-func newSyncLocal(sourceFilePath, targetFilePath string, fileSize, syncBatchSize int64, numSyncWorkers int) (*syncLocal, error) {
+func newSyncLocal(sourceFilePath, targetFilePath string, fileSize, syncBatchSize int64) (*syncLocal, error) {
 	sourceFileIo, err := NewDirectFileIoProcessor(sourceFilePath, os.O_RDONLY, 0)
 	if err != nil {
 		return nil, err
@@ -102,7 +100,6 @@ func newSyncLocal(sourceFilePath, targetFilePath string, fileSize, syncBatchSize
 		sourceFileIo:   sourceFileIo,
 		targetFileIo:   targetFileIo,
 		syncBatchSize:  syncBatchSize,
-		numSyncWorkers: numSyncWorkers,
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


sparse/local.go
```diff
@@ -0,0 +1,172 @@
...
+	})
+
+	fileInfo, err := os.Stat(sourceFilePath)
+	if err != nil {
+		log.WithError(err).Error("Failed to get file info of source file")
+		return err
+	}
+
+	fileSize := fileInfo.Size()
+	if fileSize%Blocks != 0 {
```
derekbit, line 28: nit: `Blocks` is the block size here and a bit weird. We can rename it to block size in the future.

derekbit:
Did a quick review.

What `syncContent()` does
1. Open the source file once and get a source file descriptor
2. Create the target file once and get a target file descriptor
3. Get data interval layout from the source file 
4. Initiate multiple workers (numSyncWorkers) for concurrently reading from the source file and writing to the target file.

We don't have any synchronization in the implementation. The concurrent readers and writers on a single source file descriptor and a single target file descriptor can indeed introduce race conditions.

c3y1huang:
> Did a quick review.
> 
> What `syncContent()` does
> 
> 1. Open the source file once and get a source file descriptor
> 2. Create the target file once and get a target file descriptor
> 3. Get data interval layout from the source file
> 4. Initiate multiple workers (numSyncWorkers) for concurrently reading from the source file and writing to the target file.
> 
> We don't have any synchronization in the implementation. The concurrent readers and writers on a single source file descriptor and a single target file descriptor can indeed introduce race conditions.

Wouldn't they be on different interval due to the fileIntervalChannel?

derekbit:
> Wouldn't they be on different interval due to the fileIntervalChannel?

Each worker seeks the file descriptor to a specific offset and reads or writes data, so multiple workers will race.

After discussing with @c3y1huang, we decided to use a single worker for now. We can improve it by using multiple workers in the future.

Using a single worker might not be a bad approach because:
- Sequential read and write operations usually perform well.
- We don't consume the entire disk IOPS.

cc @innobead

sparse/local.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	defer sourceFileIo.Close()
+
+	local.sourceFileIo = sourceFileIo
+
+	targetFileIo, err := NewDirectFileIoProcessor(local.targetFilePath, os.O_RDWR|os.O_CREATE, 0)
+	if err != nil {
+		return err
+	}
+	local.targetFileIo = targetFileIo
```
derekbit: Can we open the two files inside `newSyncLocal()`? Then, they can be closed in `local.close()`.

sparse/local.go
```diff
@@ -0,0 +1,169 @@
...
+func newSyncLocal(sourceFilePath, targetFilePath string, syncBatchSize int64) (*syncLocal, error) {
+	sourceFileIo, err := NewDirectFileIoProcessor(sourceFilePath, os.O_RDONLY, 0)
+	if err != nil {
+		return nil, err
+	}
+
+
+	targetFileIo, err := NewDirectFileIoProcessor(targetFilePath, os.O_RDWR|os.O_CREATE, 0)
+	if err != nil {
+		return nil, err
```
derekbit, line 87: close `sourceFileIo` if failed.

sparse/test/local_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	. "github.com/longhorn/sparse-tools/sparse"
+)
+
+func TestSyncLocalFile(t *testing.T) {
+	layoutFrom := []FileInterval{
+		{Kind: SparseData, Interval: Interval{Begin: 0, End: 1 * Blocks}},
+		{Kind: SparseHole, Interval: Interval{Begin: 1 * Blocks, End: 2 * Blocks}},
+		{Kind: SparseData, Interval: Interval{Begin: 2 * Blocks, End: 3 * Blocks}},
+	}
```
derekbit: Can we add more test cases

1. data-hole-data (you've added)
2. hole-data-hole
3. hole-data
4. data-hole

Thank you.

derekbit (approved):

PR #106, opened 2024-07-08
reset err before return nil

mschneider82:

by reviewing the transfer code for https://github.com/longhorn/longhorn/issues/8745
i have found, that the `return nil` is ineffective when the err was set before via `getLocalDiskFileChangeTimeAndChecksum()`, the defer func wraps the existing filled `err` variable and returns an error even it wants to return nil.

### DIFF
--- a/sparse/client.go
+++ b/sparse/client.go
@@ -157,6 +157,7 @@ func SyncContent(sourceName string, rw ReaderWriterAt, fileSize int64, remote st
 	if fastSync && filepath.Ext(client.sourceName) == types.SnapshotDiskSuffix {
 		if client.isLocalAndRemoteDiskFilesIdentical() {
 			log.Infof("Skipped syncing file %v", client.sourceName)
+			err = nil // the err could be set via getLocalDiskFileChangeTimeAndChecksum() and would be return via defer()
 			return nil
 		}
 	}


sparse/client.go
```diff
@@ -157,6 +157,7 @@ func SyncContent(sourceName string, rw ReaderWriterAt, fileSize int64, remote st
 	if fastSync && filepath.Ext(client.sourceName) == types.SnapshotDiskSuffix {
 		if client.isLocalAndRemoteDiskFilesIdentical() {
 			log.Infof("Skipped syncing file %v", client.sourceName)
+			err = nil // the err could be set via getLocalDiskFileChangeTimeAndChecksum() and would be return via defer()
```
derekbit, line 160: Can it move to line 149?
```
	if filepath.Ext(sourceName) == types.SnapshotDiskSuffix {
		recordedChangeTime, recordedChecksumMethod, recordedChecksum, errGetChecksum := getLocalDiskFileChangeTimeAndChecksum(sourceName)
		if errGetChecksum != nil {
			log.WithError(errGetChecksum).Warnf("Failed to get change time and checksum of local file %v", sourceName)
		}
	}
```

mschneider82:
sorry for the confussion, i have tested it via playground. it returns nil and will not be overwritten by the defer func.

https://go.dev/play/p/OhcBZQGzKoA

So this is not a bug or a fix.

derekbit:
No worries. Let's keep finding the root cause.

PR #115, opened 2025-10-09
chore(vendor): replace github.com/pkg/errors with github.com/cockroachdb/errors

derekbit:

Issue Longhorn/longhorn#11413

github.com/pkg/errors is archived and not maintained.


Dockerfile.dapper3656801351
```diff
@@ -0,0 +1,28 @@
+FROM registry.suse.com/bci/golang:1.23
```
COLDTURNIP: Temporary Dapper file accidentally committed.

scripts/entry
```diff
@@ -3,7 +3,7 @@ set -e
 
 trap "chown -R $DAPPER_UID:$DAPPER_GID ." exit
 
-export GOFLAGS=-mod=vendor
+export GOFLAGS=-mod=mod
```
COLDTURNIP, line 6: Why do we change to `mod` here?

derekbit: https://github.com/longhorn/sparse-tools/actions/runs/18381565691/job/52368997908?pr=115

COLDTURNIP (approved):

PR #120, opened 2026-02-10
add HTTP retry with exponential backoff 

mschneider82:
-Add sendHTTPRequestWithRetry with up to 5 retries and exponential backoff (2s base, 30s max) for transient server/network errors
- Apply retry logic to sendHole, getChecksum, getRecordedMetadata, and writeData requests
- Propagate context from SyncContent to syncContent for early cancellation support (including fast-sync check)
- tune HTTP transport: set ResponseHeaderTimeout, enable keep-alives, and configure 120s idle connection timeout
- Increase server idle timeout from 90s to 300s and HTTP idle timeout from 60s to 240s to prevent premature shutdown during retries

#8745 - Rebuilding Replica fails on larger volumes ( my old bug report )
#9051 - Replicas infinitely rebuilding                                                                                                                                                    

current improvment issue:                                                                                                    
#9626   https://github.com/longhorn/longhorn/issues/9626

I got stock in rebuild loop because of 3-4% packetlost again, it fails after syncing 4% sometime 9%. 

We need some retrys here for longhorn-engine (i am using v1)


sparse/client.go
```diff
@@ -157,6 +166,11 @@ func SyncContent(sourceName string, rw ReaderWriterAt, fileSize int64, remote st
 		recordedChangeTime, recordedChecksumMethod, recordedChecksum, syncBatchSize, numSyncWorkers)
 	defer client.close() // kill the server no matter success or not, best effort
 
+	// Set context early so that fast-sync check is also cancellable
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	client.ctx = ctx
```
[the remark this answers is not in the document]
mschneider82: it was created the same way on line 186 before, this is just a refactor and an improvement.

sparse/client.go
```diff
@@ -316,6 +330,58 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
+func (client *syncClient) sendHTTPRequestWithRetry(method string, action string, queries map[string]string, data []byte) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	delay := retryBaseDelay
+
+	for attempt := 0; attempt <= maxRetries; attempt++ {
+		resp, err = client.sendHTTPRequest(method, action, queries, data)
+		if err == nil && resp.StatusCode == http.StatusOK {
+			return resp, nil
+		}
```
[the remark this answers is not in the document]
mschneider82: I can fix this without changing any function signatures by modifying sendHTTPRequest (client.go:312-314) to use http.NewRequestWithContext(client.ctx, ...) instead of http.NewRequest(...).
 Since sendHTTPRequest is already a method on syncClient, it has direct access to client.ctx.

mschneider82:
I have addressed the copilot review in my newest force push.

sparse/client.go
```diff
@@ -28,6 +28,39 @@ const (
...
+		maxRetries:     maxRetries,
+		retryBaseDelay: retryBaseDelay,
+		retryMaxDelay:  retryMaxDelay,
+	}
+}
+
+// ResetTestRetryConfig resets the test retry configuration to use production defaults
+func ResetTestRetryConfig() {
+	testRetryConfig = nil
+}
```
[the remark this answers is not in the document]
mschneider82: this is fine, i want to use the same test package as the other tests.

mschneider82:
@COLDTURNIP i have addressed all points. added tests for the retrys

sparse/client.go
davidcheng0922: I think the test retry config should be moved to  _test.go file instead of being defined here.

Since it’s only used for testing, keeping it in test-only files avoids exposing test-specific APIs in production code and prevents introducing global side effects into the package.

sparse/client.go
```diff
@@ -316,6 +359,103 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
+	// Drain up to a reasonable limit to allow connection reuse without unbounded memory usage.
+	const maxDrainBytes = 1 << 20 // 1 MiB
+	_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainBytes))
+	_ = resp.Body.Close()
+}
+
+// isClientError returns true if the response is a 3xx redirect or 4xx client error
+func isClientError(resp *http.Response) bool {
+	// 3xx redirects and 4xx client errors should not be retried
+	return resp != nil && resp.StatusCode >= 300 && resp.StatusCode < 500
```
davidcheng0922: For status code 429 (`Too Many Requests`), should it also be treated as non-retry?

sparse/client.go
```diff
@@ -316,6 +359,103 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
+			default:
+			}
+		}
+		return client.ctx.Err()
+	case <-timer.C:
+		return nil
+	}
+}
+
+func (client *syncClient) sendHTTPRequestWithRetry(method string, action string, queries map[string]string, data []byte) (*http.Response, error) {
```
davidcheng0922: For the retry method, how about using an existing library like `github.com/avast/retry-go` instead of self-implementing it?

Suggest Pattern:

```go
func (m *SnapshotMonitor) waitAndHandleSnapshotHashing(engine *longhorn.Engine, engineClientProxy engineapi.EngineClientProxy,
	snapshotName string) error {
	opts := []retry.Option{
		retry.Context(m.ctx),
		retry.Attempts(snapshotHashSyncStatusAttempts),
		retry.DelayType(retry.FixedDelay),
		retry.LastErrorOnly(true),
		retry.Delay(snapshotHashSyncStatusPeriod * time.Second),
		retry.RetryIf(func(err error) bool {
			if err == nil {
				return false
			}
			return strings.Contains(err.Error(), string(engineapi.ProcessStateInProgress))
		}),
	}

	// retry does periodically fetching and syncing the snapshot hashing status.
	if err := retry.Do(func() (err error) {
		return m.syncHashStatusFromEngineReplicas(engine, engineClientProxy, snapshotName)
	}, opts...); err != nil {
		return errors.Wrapf(err, "failed to sync hash status for snapshot %v since %v", snapshotName, err)
	}

	return nil
}

```

I think it also supports exponential backoff if needed.

sparse/client.go
```diff
@@ -316,6 +359,103 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
+	cfg := defaultRetryConfig
+	if testRetryConfig != nil {
+		cfg = *testRetryConfig
+	}
```
davidcheng0922: It seems a bit odd to reference `testRetryConfig` in the non-test code path.

sparse/client.go
mschneider82: I initially implemented the retry logic without adding an external dependency to keep the codebase lean. The current implementation is straightforward, well-tested, and covers all required scenarios (exponential backoff, max retries, context cancellation).                                                                                                                        
   
However, if there's a strong preference within the Longhorn project for using `github.com/avast/retry-go`, I'm open to switching. Before we proceed with that approach, I'd like to clarify:
                  
  1. **Is this library already used in other Longhorn components?** If so, using it here would maintain consistency across the project.
  2. **Which version should we use?** The library has multiple major versions (v3, v5), and we'll need to maintain/update this dependency going forward.

  **My recommendation**: Keep the current implementation as it's simple, dependency-free, and fulfills all requirements. But I'm happy to refactor if the team prefers library consistency over minimal dependencies.

Please let me know your preference and I'll proceed accordingly.

sparse/client.go
mschneider82: thats right, have a _test file in the directory is better, i tried to use the same testing pattern like the other tests, thats why this was the only solution (or build test flag for different settings ). I will move the _test.go to the directory

sparse/client.go
mschneider82: will change that

sparse/client.go
mschneider82: Ok the tests need some helper functions from sparse/test/testutil_test.go , I have to copy them.

sparse/client.go
mschneider82: @davidcheng0922 i have added retry-go in a seperate commit, for review, please check if its needed. if its not worth adding a lib we can revert it.

sparse/client.go
```diff
@@ -155,6 +180,11 @@ func SyncContent(sourceName string, rw ReaderWriterAt, fileSize int64, remote st
 
 	client := newSyncClient(remote, sourceName, fileSize, rw, directIO, httpClientTimeout,
 		recordedChangeTime, recordedChecksumMethod, recordedChecksum, syncBatchSize, numSyncWorkers)
+
+	// Set context early so that fast-sync check is also cancellable on function return
+	ctx, cancel := context.WithCancel(context.Background())
+	client.ctx = ctx
+	defer cancel()
```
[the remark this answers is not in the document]
mschneider82: no api change here. same as before with ctx

sparse/client.go
davidcheng0922: Thank you. Next week is the Chinese Lunar New Year, so we will review the code after the holiday.

mschneider82: ok happy new year ;-) ping me if you need further changes. hope this can land in the next version

sparse/client_retry_test.go
```diff
@@ -0,0 +1,439 @@
...
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	retry "github.com/avast/retry-go/v5"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
davidcheng0922: Follow the import rules https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports

sparse/client_retry_test.go
```diff
@@ -0,0 +1,439 @@
...
+		serverAddr,
+		"test-source",
+		1024,
+		&mockReaderWriterAt{},
+		false,
+		10,
+		"", "", "",
+		512*Blocks,
+		4,
+	)
```
davidcheng0922, line 95: NIT: consider defining variables for the parameters instead of passing raw values inline

sparse/client.go
davidcheng0922: Overall LGTM — thank you for the contribution!
Let's wait for other members' comments before merging.

sparse/client_retry_test.go
mschneider82: fixed in latest commit, I also patched the server.go because it did not follow the import rules.

sparse/client_retry_test.go
mschneider82: fixed in latest commit

mschneider82:
I can rebase it into one commit if wanted.

sparse/client.go
davidcheng0922: Since other team members are using the stable `v4` of `github.com/avast/retry-go`, we should align on the same version for consistency. We can update it together in the future when needed. 

Many thanks.

mschneider82: downgrade done

sparse/client.go
derekbit, line 23:
```suggestion
import (
"bytes"
"context"
"encoding/json"
"fmt"
"io"
"net/http"
"os"
"path/filepath"
"strconv"
"sync"
"time"

"github.com/cockroachdb/errors"

retry "github.com/avast/retry-go/v4"
log "github.com/sirupsen/logrus"

"github.com/longhorn/sparse-tools/types"
"github.com/longhorn/sparse-tools/util"
)
```

sparse/client.go
```diff
@@ -316,6 +399,24 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
 
+// closeResponse drains and closes the response body
+func closeResponse(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return
+	}
+	// Drain up to a reasonable limit to allow connection reuse without unbounded memory usage.
+	const maxDrainBytes = 1 << 20 // 1 MiB
+	_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainBytes))
+	_ = resp.Body.Close()
```
derekbit, line 410: What will happen if the maxDrainBytes is not able to drain up? Should we give up thecollection reuse when it occurs?

sparse/client.go
```diff
@@ -316,6 +399,24 @@ func (client *syncClient) sendHTTPRequest(method string, action string, queries
...
+}
+
+// isNonRetryableStatusCode returns true if the HTTP status code indicates a client error
+// (3xx or 4xx) that should not be retried. 429 (Too Many Requests) is excluded since it
+// represents transient rate-limiting.
+func isNonRetryableStatusCode(statusCode int) bool {
+	return statusCode >= 300 && statusCode < 500 && statusCode != http.StatusTooManyRequests
```
derekbit, line 417: 300: http.StatusMultipleChoices
500: http.StatusInternalServerError

Can we replace them with http constants?

mschneider82: I’d keep the numeric boundaries here. We’re not matching specific status codes, we are checking a range: >= 300 && < 500 means "any 3xx/4xx". Using constants like http.StatusMultipleChoices (300) or http.StatusInternalServerError (500) can read like we’re targeting those exact codes, when they’re just range limits (and 500 is an exclusive upper bound, not an included case). I think http.StatusTooManyRequests makes sense here

sparse/client.go
mschneider82: yes, if the body exceeds 1 MiB, we won't drain it completely and the connection won't be reused. Since this function is only called for error/non-200 responses (which are typically small), losing connection reuse in the rare case of a >1MiB error response is acceptable i think.

sparse/client.go
mschneider82: I have changed the ordering in my new commit, ( your suggestion missed the go fmt )

mschneider82:
Any more comments? Should i rebase it into one commit - or will this be done in the merge ?

mschneider82:
Friendly reminder on this PR: it would be great to get this merged, as it adds retry handling with exponential backoff for transient failures and should make sync/rebuild behavior more robust. Please let me know if there are any concerns or changes needed.

sparse/client_retry_test.go
```diff
@@ -0,0 +1,453 @@
...
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	retry "github.com/avast/retry-go/v4"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
derekbit, line 16: Ref: https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports
```suggestion
	retry "github.com/avast/retry-go/v4"
	g
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
```

shuo-wu (approved):
LGTM

derekbit (approved):

derekbit:
Hello @mschneider82 
Thanks for your contribution. Could you update the go.mod of longhorn-engine? Thanks.

